PR #46, opened 2026-02-27
Add dashboard api

blehnen:
Phase 1,2,3

Phase 1 - add query handlers to transports for reading data 
Phase 2-  Create API project - all calls are read-only for now 
Phase 3 - API endpoints

Supports only the 3 DB transports at the moment - SQL server, PostGre and Sqlite

Will not be pushed to nuget until support for the other two transports is in place

Needs a full review before we continue. Due to the size of this change, lets add comments to our own PR for any issues found.


Source/DotNetWorkQueue.Dashboard.Api/Configuration/DashboardOptions.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    /// <summary>
+    /// Configuration options for the DotNetWorkQueue Dashboard API.
+    /// </summary>
+    public class DashboardOptions
+    {
+        /// <summary>
+        /// Gets or sets the route prefix for dashboard API endpoints.
+        /// </summary>
+        public string RoutePrefix { get; set; } = "api/v1/queues";
```
blehnen: Is queues the right name for the API?  On the other hand, it can be overridden in the config so I suppose it doesn't matter much.

It's also a bit confusing that we can change the prefix here for the queue controller, but the connection controller is hard coded and cannot be changed, and is also part of 'queues'.

I think this should be re-worked a bit

Source/DotNetWorkQueue.Dashboard.Api/Controllers/QueuesController.cs
```diff
@@ -0,0 +1,168 @@
...
+using DotNetWorkQueue.Dashboard.Api.Services;
+using Microsoft.AspNetCore.Mvc;
+
+namespace DotNetWorkQueue.Dashboard.Api.Controllers
+{
+    /// <summary>
+    /// API controller for per-queue dashboard operations.
+    /// </summary>
+    [ApiController]
+    [Route("api/v1/queues/queues")]
```
blehnen: The route configuration is confusing - lets revisit

Source/DotNetWorkQueue.Dashboard.Api/Services/DashboardService.cs
```diff
@@ -0,0 +1,254 @@
...
+using DotNetWorkQueue.Transport.RelationalDatabase.Basic;
+using DotNetWorkQueue.Transport.RelationalDatabase.Basic.Query;
+using DotNetWorkQueue.Transport.Shared;
+
+namespace DotNetWorkQueue.Dashboard.Api.Services
+{
+    /// <summary>
+    /// Implementation of <see cref="IDashboardService"/> that resolves query handlers from per-queue containers.
+    /// </summary>
+    internal class DashboardService : IDashboardService
```
blehnen: This class uses a service-locator anti-pattern, but due to how the rest of the solution is structured, there is no good option.  Sure, we could create a 'DashBoardDataAcces' and inject all of the queries/commands. But because we have N numbers of connections, we still need to pull that from somewhere.  We could pull them all up-front in IoC startup or use a factory but that's really just going through the motions.

Going to leave it for now, might change before the final release however.

Source/DotNetWorkQueue.Dashboard.Api/Services/IDashboardService.cs
```diff
@@ -0,0 +1,69 @@
...
+        PagedResponse<ErrorMessageResponse> GetErrors(Guid queueId, int pageIndex, int pageSize);
+
+        /// <summary>Gets error retry tracking records for a message.</summary>
+        IReadOnlyList<ErrorRetryResponse> GetErrorRetries(Guid queueId, long messageId);
+
+        /// <summary>Gets queue configuration as JSON.</summary>
+        ConfigurationResponse GetConfiguration(Guid queueId);
+
+        /// <summary>Gets all scheduled jobs for a queue.</summary>
+        IReadOnlyList<JobResponse> GetJobs(Guid queueId);
```
blehnen: This method should be deleted.

Source/DotNetWorkQueue.Dashboard.Api/DashboardApi.cs
```diff
@@ -0,0 +1,183 @@
...
+            _adminContainers.Clear();
+            _queueContainerMap.Clear();
+
+            foreach (var container in _queueContainers)
+            {
+                try
+                {
+                    container.Dispose();
+                }
+                catch (ObjectDisposedException)
```
blehnen: Did we add this for a good reason?  This isn't something we normally catch and ignore if I recall correctly.

Source/DotNetWorkQueue.Dashboard.Api/DashboardExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+                    mvcOptions.Filters.Add<DashboardExceptionFilter>();
+                })
+                .AddApplicationPart(typeof(DashboardExtensions).Assembly);
+
+            if (options.EnableSwagger)
+            {
+                services.AddEndpointsApiExplorer();
+                services.AddSwaggerGen(c =>
+                {
+                    c.SwaggerDoc("v1", new Microsoft.OpenApi.Models.OpenApiInfo
```
blehnen: Might need more options set here - double check on that

Source/DotNetWorkQueue.Dashboard.Api/Configuration/DashboardOptions.cs
blehnen: Renamed to dashboard which makes more sense. Removed prefix.  The idea was that maybe someone already has this name taken in their current API, but this could be revisited later

PR #47, opened 2026-02-28
Phase 4 

blehnen:
1. Message header and message body GET in API

TODO - Phase 4.5 - attempt to return message body as serialized json of the object. Not straight forward, the queue is very generic and relies on the user to tell it what the object is when using the queues via <T>


Source/DotNetWorkQueue.Dashboard.Api/Middleware/DashboardExceptionFilter.cs
```diff
@@ -41,6 +51,7 @@ public void OnException(ExceptionContext context)
                     break;
 
                 case InvalidOperationException:
+                    _logger.LogInformation(context.Exception, "Resource not found: {Message}", context.Exception.Message);
```
blehnen: Should this be a warning?

PR #62, opened 2026-03-11
Allow background consumer jobs (reset heartbeats, delete expired messages, etc) to be handled elsewhere

blehnen:
Add MaintenanceMode to decouple maintenance monitors from consumers

Introduce MaintenanceMode (Consumer/External) on QueueConsumerConfiguration
so consumers can opt out of running maintenance monitors (heartbeat reset,
expiration cleanup, error cleanup) when those tasks run elsewhere.

 Add IQueueMaintenanceService/QueueMaintenanceService as a standalone wrapper
  around the transport's IQueueMonitor. Wire into the Dashboard API with a
  per-queue HostMaintenance option that eagerly starts monitors at startup
  and exposes status via GET /api/dashboard/queues/{id}/maintenance.

### DIFF
--- a/Source/DotNetWorkQueue/Queue/QueueMaintenanceService.cs
+++ b/Source/DotNetWorkQueue/Queue/QueueMaintenanceService.cs
@@ -79,7 +79,7 @@ namespace DotNetWorkQueue.Queue
         public bool IsRunning => _started && !IsDisposed;
 
         /// <inheritdoc />
-        public DateTime? LastRun => null; // TODO: wire up when BaseMonitor exposes last-run timestamps
+        public DateTime? LastRun => _queueMonitor.LastRunUtc;
 
         #region IDisposable, IIsDisposed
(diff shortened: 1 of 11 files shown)


Source/DotNetWorkQueue/Queue/QueueMaintenanceService.cs
```diff
@@ -0,0 +1,130 @@
...
+            _queueMonitor.Stop();
+            _started = false;
+            _log.LogInformation("Queue maintenance service stopped");
+        }
+
+        /// <inheritdoc />
+        public bool IsRunning => _started && !IsDisposed;
+
+        /// <inheritdoc />
+        public DateTime? LastRun => null; // TODO: wire up when BaseMonitor exposes last-run timestamps
```
blehnen: Whoops - I missed a TODO; Lets fix this

PR #137, opened 2026-04-22
fix(transports): options factory re-loads after default fallback (#120)

### LINKED ISSUE #120: TransportOptionsFactory permanently caches default options when queue doesn't exist at first resolution
## Problem

`PostgreSqlMessageQueueTransportOptionsFactory.Create()` (and peer factories for SqlServer, SQLite, LiteDb, Redis) lazily load transport options from the DB/store and cache the result. When the queue's configuration table/record doesn't exist at first resolution, the factory falls through to a fresh `new PostgreSqlMessageQueueTransportOptions()` with all defaults, and caches that default for the lifetime of the container. There is **no refresh / re-load mechanism** if the queue is subsequently created.

Any consumer of `IBaseTransportOptions` resolved through this factory is stuck on defaults, even after the queue comes into existence. This affects every option on the transport — `EnableHistory`, `EnableStatus`, `EnableHeartBeat`, `EnableDelayedProcessing`, `EnableRoute`, and so on.

## Reproduction

- Start the dashboard container with a connection string targeting a queue that doesn't exist yet.
- Anything in the dashboard-admin container that resolves `IBaseTransportOptions` (e.g., maintenance service startup, various decorator chains) now has defaults cached.
- Create the queue with non-default options via the producer/consumer app.
- The dashboard still sees defaults — the factory will not reload.

Shipped workaround right now: restart the dashboard after queue creation.

## Surfaced via #119 (history reads)

Fix in [PR #119](https://github.com/blehnen/DotNetWorkQueue/pull/119) removes the `EnableHistory` read-side dependency, which fixes the user-visible history bug without resolving this root-cause. Other options remain affected (though they're less frequently consumed from the dashboard side).

## Affected files

- `Source/DotNetWorkQueue.Transport.PostgreSQL/Basic/Factory/PostgreSqlMessageQueueTransportOptionsFactory.cs`
- `Source/DotNetWorkQueue.Transport.SqlServer/Basic/Factory/SqlServerMessageQueueTransportOptionsFactory.cs`
- `Source/DotNetWorkQueue.Transport.SQLite/Basic/Factory/SqLiteMessageQueueTransportOptionsFactory.cs`
- `Source/DotNetWorkQueue.Transport.LiteDB/Basic/Factory/LiteDbMessageQueueTransportOptionsFactory.cs`
- `Source/DotNetWorkQueue.Transport.Redis/Basic/Factory/RedisMessageQueueTransportOptionsFactory.cs`

(all have the same cache-on-null pattern — one line near `if (_options == null) _options = new ...()`)

## Proposed approaches

1. **Don't cache the default-fallback** — retry loading on each `Create()` call when the previous attempt returned defaults. Costs one DB/store lookup per resolve; acceptable if resolution frequency is low. The happy path (options loaded successfully) still caches.
2. **Time-boxed cache** — cache loaded options with a TTL; re-load on expiry. More complex; risks staleness windows but avoids per-call DB hits.
3. **Event-driven invalidation** — add a "config table now exists" signal (e.g., cache invalidation when the admin container sees a newly appearing table). Highest complexity.

Lean: **option 1** — it's the smallest, matches "correct and complete" preference, and the extra DB lookup only happens while options are defaults (transient state).

## Acceptance

- Integration test: start a container against a non-existent queue, resolve `IBaseTransportOptions` once, create the queue with non-default options, re-resolve — asserts the factory now returns the non-default options.
- All five transport option factories covered.

blehnen:
Closes #120.

## Summary
Before: when `*MessageQueueTransportOptionsFactory.Create()` resolved against a non-existent queue, the fresh-default options instance it returned was cached on `_options` and stuck for the lifetime of the container. Any subsequent `Create()` after queue creation still saw defaults.

After: the fresh-default fallback is returned **without** being cached. `_options` stays null until a real load succeeds, so a subsequent `Create()` re-attempts the store lookup and observes the newly-persisted options. Happy-path caching is unchanged — one load per container lifetime once real data exists.

## Scope
5 factories patched; same shape, one-line semantic change per file:

| Transport | File | Notes |
|---|---|---|
| PostgreSQL | `Source/DotNetWorkQueue.Transport.PostgreSQL/Basic/Factory/PostgreSqlMessageQueueTransportOptionsFactory.cs` | Canonical |
| SqlServer | `Source/DotNetWorkQueue.Transport.SqlServer/Basic/Factory/SQLServerMessageQueueTransportOptionsFactory.cs` | Capital SQL in filename |
| SQLite | `Source/DotNetWorkQueue.Transport.SQLite/Basic/Factory/SqLiteMessageQueueTransportOptionsFactory.cs` | |
| LiteDb | `Source/DotNetWorkQueue.Transport.LiteDB/Basic/Factory/LiteDbMessageQueueTransportOptionsFactory.cs` | **InMemoryOptionsCache fallback preserved** (hit still caches — it's real persisted data) |
| Redis | `Source/DotNetWorkQueue.Transport.Redis/Basic/RedisTransportOptionsFactory.cs` | Different file name + shape (reads `StringGet` on `_redisNames.Configuration`); added `protected virtual GetDb()` seam following the existing Redis-handler pattern |

## Tests
- **PostgreSQL**: 4 tests — null-load-doesn't-cache, happy-path-caches, after-defaults-observes-new, preserved smoke test
- **SqlServer**: same 4
- **SQLite**: same 4
- **LiteDb**: 4 — three core + `InMemoryOptionsCache hit caches and sticks`
- **Redis**: 3 core — key-missing-doesn't-cache, key-has-json-caches, after-defaults-observes-new (uses `TestableFactory` with `GetDb()` override)

**Regression (full suites, all green):**
- PostgreSQL **117/117**
- SqlServer **130/130**
- SQLite **129/129**
- LiteDb **170/170**
- Redis **186/186**
- Total **732**

## Deferred
- **End-to-end integration test** (plan Task 6). The per-factory unit tests pin down the cache-skip behavior at the correct seam. Cross-container timing validation would need significant scaffolding (two independent `QueueContainer`s sharing a file-mode SQLite DB with careful ordering) — deferring to a follow-up once the unit-level fix is merged. Recommendation: smoke-verify in Dashboard.Ui against a real transport.

## Test plan
- [x] Per-transport unit tests (20+ new, all green)
- [x] Full unit-test suites on all 5 affected transports (732 tests green)
- [ ] Jenkins PR build — 14 parallel integration stages
- [ ] Post-merge: Dashboard.Ui smoke-test against a pre-existing queue scenario

### DIFF
--- a/Source/DotNetWorkQueue.Transport.LiteDB/Basic/Factory/LiteDbMessageQueueTransportOptionsFactory.cs
+++ b/Source/DotNetWorkQueue.Transport.LiteDB/Basic/Factory/LiteDbMessageQueueTransportOptionsFactory.cs
@@ -36,6 +36,7 @@ namespace DotNetWorkQueue.Transport.LiteDb.Basic.Factory
         private readonly IConnectionInformation _connectionInformation;
         private readonly object _creator = new object();
         private LiteDbMessageQueueTransportOptions _options;
+        private bool _loadedFromStore;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="LiteDbMessageQueueTransportOptionsFactory"/> class.
@@ -63,30 +64,37 @@ namespace DotNetWorkQueue.Transport.LiteDb.Basic.Factory
                 return new LiteDbMessageQueueTransportOptions();
             }
 
-            if (_options != null) return _options;
+            if (_loadedFromStore) return _options;
             lock (_creator)
             {
-                if (_options != null) return _options;
+                if (_loadedFromStore) return _options;
 
                 var loaded = _queryOptions.Handle(new GetQueueOptionsQuery<LiteDbMessageQueueTransportOptions>());
                 if (loaded != null)
                 {
                     _options = loaded;
+                    _loadedFromStore = true;
                     return _options;
                 }
 
                 // Fallback: static cache for in-memory mode (producer/consumer with separate DB instances).
-                // A cache hit here IS real persisted data, so we cache it on the instance field.
+                // A cache hit here IS real persisted data, so we lock it in as the loaded value.
                 var key = $"{_connectionInformation.QueueName}|{_connectionInformation.ConnectionString}";
                 if (InMemoryOptionsCache.TryGetValue(key, out var cached))
                 {
                     _options = cached;
+                    _loadedFromStore = true;
                     return _options;
                 }
 
-                // Not in store, not in static cache — return defaults but do NOT cache.
-                // A subsequent Create() after queue creation must observe the newly-persisted options.
-                return new LiteDbMessageQueueTransportOptions();
+                // Not in store, not in static cache — return a tentative default
+                // whose reference is stable across calls, so callers that mutate the
+                // returned instance (e.g. via the Creation class's Options property)
+                // see their mutations persist. We re-query the store on every Create()
+                // call until it returns non-null, at which point we swap the cached
+                // reference to the loaded options.
+                if (_options == null) _options = new LiteDbMessageQueueTransportOptions();
+                return _options;
             }
         }
 
--- a/Source/DotNetWorkQueue.Transport.LiteDB/Basic/Factory/LiteDbMessageQueueTransportOptionsFactory.cs
+++ b/Source/DotNetWorkQueue.Transport.LiteDB/Basic/Factory/LiteDbMessageQueueTransportOptionsFactory.cs
@@ -66,22 +66,28 @@ namespace DotNetWorkQueue.Transport.LiteDb.Basic.Factory
             if (_options != null) return _options;
             lock (_creator)
             {
-                if (_options == null)
+                if (_options != null) return _options;
+
+                var loaded = _queryOptions.Handle(new GetQueueOptionsQuery<LiteDbMessageQueueTransportOptions>());
+                if (loaded != null)
                 {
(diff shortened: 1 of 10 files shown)


blehnen:
Reverting this approach. The "re-query the store on every `Create()` call while tentative" strategy turned a single cached-default read into N concurrent reads during the queue-creation window (setOptions, CreateQueueTables handler, producer init, consumer init, heartbeat monitor, etc. all resolve `factory.Create()` back-to-back).

Those reads go through `GetQueueOptionsQueryHandler → GetTableExistsQuery → sqlite_master` (SQLite) or the equivalent SQL Server catalog lookup, which collides with the concurrent `CREATE TABLE` from `CreateQueueTablesAndSaveConfigurationCommandHandler`. SQLite surfaces it as `database schema is locked: main`; SQL Server as `Execution Timeout Expired`. Polly's retry decorator then backs off (3.9s → 17s → 21s exponential) and the cascade compounds into multi-hour Jenkins stalls (observed on multiple integration test stages; master is unaffected).

Leaving issue #120 open — a future fix should avoid polling the store on every `Create()`. Likely approach: explicit invalidation signal from `CreateQueueTablesAndSaveConfigurationCommandHandler` (or the SQL equivalent) after persist, so the factory re-reads exactly once at the moment the queue transitions from "doesn't exist" to "exists".

PR #181, opened 2026-07-06
Refactor DataStorage read-lock boilerplate into ReadLockScope (clears 16 S2222 FPs)

blehnen:
Resolves the 16 `csharpsquid:S2222` findings in `Transport/Memory/Basic/DataStorage.cs` — which were **false positives**, not lock leaks — by removing the duplicated boilerplate that confused the analyzer.

## Background
All 16 read methods repeated the same idiom:
```csharp
var tookLock = false;
try
{
    if (!_lock.IsReadLockHeld) { _lock.EnterReadLock(); tookLock = true; }  // re-entrancy guard
    ...
}
finally { if (tookLock) _lock.ExitReadLock(); }                            // released on every path
```
The lock **is** released on all paths; `_lock` is a non-recursive `ReaderWriterLockSlim`, so the `IsReadLockHeld` guard makes the outer frame own the lock and inner frames no-op. S2222's dataflow can't model the `tookLock` flag, so it false-flagged all 16. (Tell: `Clear()`'s plain `try/finally` write lock was never flagged.)

## Change
- Add `private readonly struct ReadLockScope : IDisposable` encapsulating the guard + guaranteed release.
- Convert the 16 methods to `using (new ReadLockScope(_lock)) { ... }`.
- Method bodies are byte-identical; only the boilerplate changed. `Clear()`'s write lock left as-is.

Same semantics, allocation-free (readonly struct), and 16 copies of subtle lock code collapse to one.

## Verification
- Release build (core project, `TreatWarningsAsErrors`): **0 warnings, 0 errors**
- Memory unit tests: **39/39**
- Memory integration tests (full producer/consumer flow through DataStorage): **57/57**

## Note
If SonarCloud still flags S2222 once on the struct's ctor (enter in ctor / release in `Dispose`), I'll add a single justified suppression there — down from 16 to at most 1, with cleaner code either way.

## Summary by CodeRabbit

* **Refactor**
  * Improved thread-safe read-lock handling across message, job, and queue read operations for more consistent behavior.
  * Simplified the internal locking approach in read paths to reduce lock bookkeeping and improve reliability under load.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -52,6 +52,10 @@ jobs:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
         shell: powershell
         run: |
-          ${{ runner.temp }}\scanner\dotnet-sonarscanner begin /k:"blehnen_DotNetWorkQueue" /o:"blehnen" /d:sonar.exclusions="**/*.png,**/*.ico,**/*.jpg,**/*.jpeg,**/*.gif,**/*.webp,**/*.woff,**/*.woff2,**/*.ttf,**/*.eot"
+          # Code coverage is measured and enforced via Codecov (Coverlet on the Jenkins
+          # integration pipeline), not SonarCloud. sonar.coverage.exclusions=** tells Sonar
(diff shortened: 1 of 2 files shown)


blehnen:
@coderabbitai full review

PR #191, opened 2026-07-09
refactor: guard conditional log calls with IsEnabled (CA1873)

blehnen:
## Summary
Bucket B part 2 of the SonarCloud code-smell reduction effort. Wraps all **39 CA1873** sites ("evaluation of argument may be expensive if logging is disabled") in an `IsEnabled(level)` guard.

All flagged sites are low-severity levels — **18 Debug, 7 Trace, 14 Information** (zero Warning/Error/Critical) — so guarding is a genuine win: the `params object[]` allocation and property/method-call argument evaluation are now skipped when the level is off. Several sites are in the continuously-running worker / heartbeat / scheduler loops.

```csharp
// before
Log.LogDebug("[{Name}] server difference is {OffsetMs} MS", Name, Offset.TotalMilliseconds);
// after
if (Log.IsEnabled(LogLevel.Debug))
    Log.LogDebug("[{Name}] server difference is {OffsetMs} MS", Name, Offset.TotalMilliseconds);
```

## Context
Complements #190 (merged), which cleared **CA2254 / S2629** by converting interpolated log messages to constant templates. CA1873 is orthogonal to that templating and was not affected — it required these guards.

## Scope
- 39 guards across 21 files (core queue/scheduler/heartbeat, Dashboard API/UI, LiteDB + Redis transports, integration-test notification helpers, and the new log-template test).
- No behavior change: guarded calls only skip when the level is disabled.

## Verification
- Full solution builds clean (0 errors; pre-existing TripleDES/SYSLIB warnings only).
- `LogMessageTemplateTests` 5/5 pass.

_Note: mechanical PRs continue to trip the duplication quality gate (deferred repo-wide dedup pass); expected._

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary log output by emitting debug/trace/info messages only when the corresponding log level is enabled across background services, workers, scheduling, health monitoring, and storage/transport checks.
  * Improved logging-related test behavior by guarding log calls based on the enabled log level, while keeping existing message rendering assertions intact.

### DIFF
--- a/Source/DotNetWorkQueue.Dashboard.Api/DashboardApi.cs
+++ b/Source/DotNetWorkQueue.Dashboard.Api/DashboardApi.cs
@@ -157,9 +157,10 @@ namespace DotNetWorkQueue.Dashboard.Api
                     var service = container.GetInstance<IQueueMaintenanceService>();
                     service.Start();
                     _maintenanceServices[kvp.Key] = service;
-                    _logger.LogInformation(
-                        "Started maintenance service for queue {QueueName} ({QueueId})",
-                        kvp.Value.QueueName, kvp.Key);
+                    if (_logger.IsEnabled(LogLevel.Information))
+                        _logger.LogInformation(
+                            "Started maintenance service for queue {QueueName} ({QueueId})",
+                            kvp.Value.QueueName, kvp.Key);
                 }
                 catch (Exception ex)
                 {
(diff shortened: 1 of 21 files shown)


blehnen:
Re: the LGPL-2.1 header nit on `CreateNotifications.cs` — declining as out of scope. That file is in `DotNetWorkQueue.IntegrationTests.Shared`, a test-support project where only **1 of 88** files carries the header (the wider test suite follows the same practice — e.g. `DotNetWorkQueue.Tests` is 5/175). Library/production projects are ~100% headered and are where the convention applies. Adding the header to just this one file would make it a lone 2/88 inconsistency on a PR scoped to CA1873. If we want headers across the test projects, that's a dedicated pass, not this PR.

PR #194, opened 2026-07-10
refactor: clear ~128 mechanical SonarCloud code smells (Track 2)

blehnen:
## Track 2 — mechanical SonarCloud code-smell cleanup

Follow-up to #190/#191/#192/#193. Pure quality pass: **128 real, behavior-preserving fixes**, plus **57 sites deliberately left untouched** because they'd be unsafe on a published library (documented below for a maintainer/profile decision).

### Fixed (128)

| Rule | n | Change |
|------|---|--------|
| S3261 | 23 | Delete dead empty-namespace stub files (types were moved to `Transport.Shared`) |
| S1905 | 31 | Remove redundant casts — `((DbCommand)command)`, `(long)db.*Length()`, `(SqlConnection)`/`(NpgsqlConnection)` on already-typed members |
| S1066 | 31 | Merge collapsible nested `if` statements into `&&` |
| S108 | 18 | Document/remove empty blocks — no-op NoOp setters, empty dispose-count guards, best-effort UI `catch` blocks; SQLite decorators use `File.Create(..).Dispose()` |
| S927 | 24 | Rename parameters to match the interface declaration (e.g. `itemData`→`property`, `enableRoute`→`enableRoutes`, `actionToRun`→`job`) |
| CA1822/S2325 | 1 | Make the private `MessageMethodHandling.HandleRawAction` static |

### Intentionally NOT changed (57) — needs a maintainer/profile decision

These are real Sonar findings, but the mechanical "fix" would change public API or runtime behavior on a published NuGet library:

- **CA1822/S2325 ×35** — the remaining make-static hits are **public/protected** members. Converting an instance member to static is a source-breaking API change. Better suppressed for public members in the quality profile.
- **S1168 ×14** — the flagged `return null`s are **load-bearing sentinels**, not empty collections: "header/message not found", "connection disposed", "no user params configured". `ReceiveMessage.cs` even comments `//NOTE - could be null` and branches query-cache logic on it. Converting `null`→empty is a behavior change requiring per-caller review.
- **S1006 ×8** — the interfaces declare the param with **no default**; the impls add `= null`. Removing the impl default is source-breaking (callers do `queue.Start()`); adding a matching default to the public interface is a deliberate API change (makes the param optional in the contract). Either direction is a maintainer call.

### Verification
- `DotNetWorkQueueNoTests.sln` builds clean: **0 warnings / 0 errors**
- Tests pass: **920** core + **57** Memory-integration + **247** RelationalDatabase (0 failures). DataStorage.cs (heaviest change surface) is exercised by the Memory integration suite.

## Summary by CodeRabbit

- **Refactor**
  - Simplified queue, message, transaction, disposal, and database operations while preserving existing behavior.
  - Improved in-memory queue state handling for deleting, finding, clearing, and tracking messages.
  - Consolidated shared dashboard transport components for consistent access across supported storage systems.
  - Streamlined asynchronous database and transaction processing.

- **Bug Fixes**
  - Preserved safer handling of expired messages, stale consumers, rollbacks, and cancellation scenarios.
  - Clarified best-effort error handling in dashboard detail and refresh operations.

- **Documentation**
  - Added inline explanations for intentionally ignored exceptions and no-op disposal paths.


blehnen:
@coderabbitai Reverted all S927 parameter renames in a5873e55.

For the record: the interfaces (`IJobScheduler`, `IAdditionalMessageData`, `IMessage`, etc.) already declare the canonical names (`job`, `property`, `queueCreator`), so the change only aligned the concrete implementations and the public *interface* contract was unchanged. But since the concrete classes are public and carry the parameter names too, reverting keeps this PR strictly non-breaking for concrete-type named-argument callers. The impl/interface param-name mismatch (S927) is left as a maintainer decision.

Remaining changes are behavior-preserving: dead empty-namespace file deletions (S3261), redundant cast removal (S1905), collapsible-if merges (S1066), empty-block documentation (S108), and one private static conversion (CA1822).

blehnen:
### Rule keys for the SonarCloud disposition pass

For the findings intentionally left unfixed here (published-API safety), to clear them from the badge without code changes:

| Rule key | n | What | Suggested disposition |
|---|---|---|---|
| \`csharpsquid:S927\` | 24 | impl param names don't match interface | **Accept** (keep rule live for new code) |
| \`external_roslyn:CA1822\` + \`csharpsquid:S2325\` | 35 | make public/protected member static → API break | **Accept** or profile-disable for public members |
| \`csharpsquid:S1006\` | 8 | override adds a default the interface lacks | **Accept** or profile-disable |

Leave **\`csharpsquid:S1168\`** (14, "return empty not null") *active* — it's a legitimate rule; ours are just intentional null-sentinels, so **Accept** those 14 individually rather than disabling the rule.

Intrinsic rules better disabled in the profile (design choices, not defects): \`csharpsquid:S107\` (params), \`csharpsquid:S2925\` (test Thread.Sleep), \`csharpsquid:S3776\` (cognitive complexity).

PR #200, opened 2026-07-11
Harden SNTP time client + tests; RedisNames init guard (#199, #185)

### LINKED ISSUE #185: RedisNames.BuildListIfNeeded uses a fragile hardcoded count guard
## Summary

`RedisNames.BuildListIfNeeded()` guards initialization with a hardcoded dictionary count:

```csharp
private void BuildListIfNeeded()
{
    if (_names.Count == 17) return; //don't return unless all names are loaded
    lock (_names)
    {
        if (_names.Count != 0) return;
        _names.Add("Notification", ...);
        // ... 17 entries total
    }
}
```

The magic number `17` must be kept in sync by hand with the number of `_names.Add(...)` calls. If anyone adds (or removes) a key name, the guard silently breaks: too-high a target means the fast-path check never short-circuits; too-low means it can return before all names load.

This is **correct today** — it's a maintainability/fragility issue, not an active bug.

## Location

`Source/DotNetWorkQueue.Transport.Redis/Basic/RedisNames.cs` (~line 31 constant / `BuildListIfNeeded`)

## Suggested fix

Derive the expected count instead of hardcoding it — e.g. gate on an `_initialized` flag set inside the lock, or use `Lazy<T>`/`LazyInitializer` so the "all names loaded" condition can't drift from the number of entries.

## Notes

Found during review of the Redis Lua script usage. Low priority — flag for a future cleanup pass.

### LINKED ISSUE #199: Harden SNTP time-client integration tests against public NTP flakiness (GuerrillaNtp socket timeouts)
## Symptom

Intermittent CI failure (seen repeatedly on the Jenkins agent), e.g. on PR #198:

```
Failed GetCurrentUnixTimestampMilliseconds_Returns_Value [8 s]
System.Net.Sockets.SocketException: Connection timed out
   at System.Net.Sockets.Socket.Receive(...)
   at GuerrillaNtp.NtpClient.Query()
   at DotNetWorkQueue.Transport.Redis.Basic.Time.SntpUnixTime.GetUnixTime() SntpUnixTime.cs:line 67
   at ...SntpUnixTimeTests.GetCurrentUnixTimestampMilliseconds_Returns_Value() SntpUnixTimeTests.cs:line 16
```

## Root cause

This is **not** a general test-helper / transport issue — it is specific to the SNTP time client:

1. **Public NTP pool over UDP.** `SntpTimeConfiguration` defaults to `Server = "pool.ntp.org"` with an **8s timeout**. `SntpUnixTime.GetUnixTime()` makes a single, un-retried `_ntpClient.Query()` UDP call. A dropped/ignored UDP reply = 8s socket timeout = test failure.
2. **Burst of fresh clients, no shared cache.** `SntpUnixTimeTests` has **6 test methods**, each calling `Create()` → `new SntpUnixTime(...)` with an empty offset cache. So each test issues its own live NTP query — **6 queries in quick succession** from one source IP (more with `--retry-failed-tests` and on every PR build). Public NTP pools rate-limit / silently drop bursts, which presents exactly as a UDP timeout.
3. **No retry anywhere** — not in production (`SntpUnixTime.GetUnixTime()`) nor in the tests.

**Why the transport itself never flakes:** the Redis transport default is `TimeServer = TimeLocations.RedisServer` (`RedisQueueInit.cs`), i.e. it uses the Redis `TIME` command, not NTP. Only these 6 SNTP-specific tests reach `pool.ntp.org`. The general produce/consume test helpers are unaffected because they never touch NTP — not because they retry.

The underlying network unreliability is on the CI host and can't be fixed short-term, so the code should tolerate transient NTP loss.

## Proposed hardening (in priority order)

1. **Share one `SntpUnixTime` across the 6 tests** via `[ClassInitialize]`/class fixture so the offset is fetched **once** and the existing `TimeExpired()` cache serves the rest — turns 6 network queries into 1, killing the burst.
2. **Add a Polly retry in production `SntpUnixTime.GetUnixTime()`** around `_ntpClient.Query()` — 2–3 attempts with **jittered** backoff (the library already depends on Polly 8.6.5). Dual-purpose: hardens real NTP usage and the tests. Jitter is important: immediate retries against a rate-limiting pool get dropped too. Decide fallback on total failure (rethrow vs. fall back to local machine time).
3. **Add `[TestCategory("ExternalNetwork")]`** to the SNTP tests so a flaky host can exclude them (mirrors the existing `StarvationBaseline` exclusion) without dropping coverage elsewhere.

## Non-goals / clarification

- Do **not** blanket-wrap the general Redis integration-test helpers in a retry — they use Redis server time and are not the source of this flakiness.
- Consider whether the SNTP tests should tolerate a hard-offline environment (assert-or-inconclusive) vs. simply being retried — an integration test that requires NTP arguably *should* fail when NTP is unreachable, so the category escape hatch (3) may be the right lever there.

## Affected files
- `Source/DotNetWorkQueue.Transport.Redis/Basic/Time/SntpUnixTime.cs`
- `Source/DotNetWorkQueue.Transport.Redis/Basic/Time/SntpTimeConfiguration.cs`
- `Source/DotNetWorkQueue.Transport.Redis.IntegrationTests/Basic/Time/SntpUnixTimeTests.cs`

blehnen:
Closes #199. Closes #185.

Two small, cohesive Redis-transport hardening fixes bundled into one PR (one serialized Jenkins build instead of two).

## #199 — SNTP integration-test NTP flakiness

Intermittent Jenkins failures: `SntpUnixTimeTests.*` → `SocketException: Connection timed out` from `GuerrillaNtp.NtpClient.Query()`. Root cause is NTP-specific, not a general transport/helper issue — the transport defaults to `TimeLocations.RedisServer` (Redis `TIME` command) and never touches NTP, so only these 6 SNTP tests reach `pool.ntp.org`. Each test built a fresh `SntpUnixTime` (empty cache) → a **burst of 6 live UDP queries** from one IP, which public pools rate-limit/drop; there was no retry anywhere.

Fixes (issue's priority order):
1. **Share one `SntpUnixTime`** across the 6 tests via `[ClassInitialize]` — offset fetched **once**, the client's `TimeExpired()` cache serves the rest. Local run: 6 tests in **407 ms** (was ~6 separate queries).
2. **Jittered Polly retry** (3 attempts, exponential + jitter) around `_ntpClient.Query()` in production `GetUnixTime()`, matching the existing SQLite/SQL retry idiom. Hardens real NTP usage too. On total failure the last exception still propagates — a genuinely unreachable NTP source *should* surface, which is what (3) is for.
3. **`[TestCategory("ExternalNetwork")]`** on the tests so a flaky host can exclude them via `--filter "TestCategory!=ExternalNetwork"` (mirrors the existing `StarvationBaseline` exclusion) without dropping coverage elsewhere.

## #185 — RedisNames fragile count guard

`BuildListIfNeeded()` gated on a hardcoded `_names.Count == 17`, which had to be hand-synced with the number of `_names.Add(...)` calls. Replaced with a `volatile bool _namesLoaded` flag set **after** all names are added, so the guard can't drift when key names are added/removed (also avoids reading `Dictionary.Count` outside the lock).

## Verification
- `Release -p:CI=true` build of the Redis transport: 0 warnings / 0 errors (XML-doc + warnings-as-errors clean).
- Redis unit tests: 187/187 pass (incl. `RedisNamesTests`, which covers the full name set).
- SNTP integration tests: 6/6 pass locally.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of network time retrieval by retrying transient NTP failures with exponential backoff and jitter.
  - Added warning logs during retry attempts to make time-sync issues easier to diagnose.
  - Enhanced concurrent initialization for Redis name mappings to avoid redundant rebuilds and reduce contention.

- **Tests**
  - Updated external-network integration tests to reuse a shared cached time source and labeled the tests as requiring outbound UDP access.

### DIFF
--- a/Source/DotNetWorkQueue.Transport.Redis.IntegrationTests/Basic/Time/SntpUnixTimeTests.cs
+++ b/Source/DotNetWorkQueue.Transport.Redis.IntegrationTests/Basic/Time/SntpUnixTimeTests.cs
@@ -6,22 +6,41 @@ using NSubstitute;
 
 namespace DotNetWorkQueue.Transport.Redis.Integration.Tests.Basic.Time
 {
+    /// <summary>
+    /// Exercises the SNTP time client against a live NTP server (default pool.ntp.org).
+    /// </summary>
+    /// <remarks>
+    /// These tests reach the public NTP pool over UDP, so they depend on outbound network access.
+    /// A single <see cref="SntpUnixTime"/> instance is shared across every test via <see cref="ClassInit"/>
+    /// so the offset is fetched once (the client caches it) instead of issuing a burst of queries that a
+    /// public pool may rate-limit or drop. Tagged <c>ExternalNetwork</c> so a host with unreliable outbound
+    /// NTP can exclude them via <c>--filter "TestCategory!=ExternalNetwork"</c>.
+    /// </remarks>
     [TestClass]
+    [TestCategory("ExternalNetwork")]
     public class SntpUnixTimeTests
     {
+        private static SntpUnixTime _test;
+
+        [ClassInitialize]
+        public static void ClassInit(TestContext context)
+        {
+            var log = Substitute.For<ILogger>();
+            var configuration = new SntpTimeConfiguration();
+            _test = new SntpUnixTime(log, configuration);
+        }
+
         [TestMethod]
         public void GetCurrentUnixTimestampMilliseconds_Returns_Value()
         {
-            var test = Create();
-            var result = test.GetCurrentUnixTimestampMilliseconds();
+            var result = _test.GetCurrentUnixTimestampMilliseconds();
             Assert.IsGreaterThan(0, result);
         }
 
         [TestMethod]
         public void GetCurrentUnixTimestampMilliseconds_Is_Recent()
         {
-            var test = Create();
-            var result = test.GetCurrentUnixTimestampMilliseconds();
+            var result = _test.GetCurrentUnixTimestampMilliseconds();
 
             var epoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
             var localUnixMs = (long)(DateTime.UtcNow - epoch).TotalMilliseconds;
@@ -34,9 +53,8 @@ namespace DotNetWorkQueue.Transport.Redis.Integration.Tests.Basic.Time
         [TestMethod]
         public void GetCurrentUnixTimestampMilliseconds_Cached_On_Second_Call()
         {
-            var test = Create();
-            var first = test.GetCurrentUnixTimestampMilliseconds();
-            var second = test.GetCurrentUnixTimestampMilliseconds();
+            var first = _test.GetCurrentUnixTimestampMilliseconds();
+            var second = _test.GetCurrentUnixTimestampMilliseconds();
 
             // Second call should use cached offset, so results should be very close
             Assert.IsLessThan(1000, Math.Abs(second - first),
@@ -46,10 +64,9 @@ namespace DotNetWorkQueue.Transport.Redis.Integration.Tests.Basic.Time
         [TestMethod]
         public void GetAddDifferenceMilliseconds_Adds_Offset()
         {
-            var test = Create();
-            var current = test.GetCurrentUnixTimestampMilliseconds();
+            var current = _test.GetCurrentUnixTimestampMilliseconds();
             var difference = TimeSpan.FromMinutes(5);
-            var result = test.GetAddDifferenceMilliseconds(difference);
+            var result = _test.GetAddDifferenceMilliseconds(difference);
 
             var expectedDifferenceMs = (long)difference.TotalMilliseconds;
             // Allow small tolerance for elapsed time between calls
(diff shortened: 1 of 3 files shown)


blehnen:
Addressed the NTP-refresh lock-hold concern in 4ec19b85.

The core observation is right: adding the retry extended the worst-case time the query holds `_getTime` (up to ~34s if the NTP server is unreachable), and every other caller would block behind it.

I didn't take the "move the query fully outside the lock" fix, though, because it reintroduces the problem this PR exists to solve: with the query outside the lock, *every* thread that races past `TimeExpired()` fires its own concurrent NTP query — the exact burst that public pools rate-limit (#199). The lock is deliberately serializing refreshes.

Instead I kept refreshes serialized but stopped callers from blocking on the slow one, via `Monitor.TryEnter`:
- **Cold start** (no cached offset yet) → `Monitor.Enter`, wait for the one query.
- **Warm cache, expired** → one thread wins `TryEnter` and refreshes; all others return the last-known offset immediately (no block, no extra queries).

So only a single thread ever queries, and no caller blocks for the retry duration. Serving a few-seconds-stale offset during an in-flight refresh is harmless for a value that's cached for the configured refresh window anyway.

PR #211, opened 2026-07-21
Coverage phase 1: JobScheduler unit tests + exclude Blazor Program bootstrap

blehnen:
## Phase 1 — Cheap structural wins (coverage recovery)

Second phase of the 90% coverage recovery (follows #210). **No production behavior changes** — 73 new unit tests plus one attribute-only file.

### Changes
- **`ProgramCoverage.cs`** (new) — attaches `[ExcludeFromCodeCoverage]` to the compiler-generated top-level-statements `Program` class, dropping Blazor host bootstrap (218 lines @ 0%) from the coverable denominator. Declared `internal` to match the accessibility the compiler synthesizes, so the assembly surface is unchanged.
- **`PendingEventHeapTests.cs`** — 20 tests: empty-heap guards, sift-up, grow-and-copy past the initial capacity, sift-down across insertion orders, duplicate timestamps, non-mutating peek, interleaved push/pop.
- **`ScheduledJobTests.cs`** — 24 tests: start/stop/update-schedule incl. all missed-event-window branches; `RunPendingEventAsync` across stale run id, queued, requeued, already-queued, send-reports-exception, send-throws, and the unusable-next-occurrence path.
- **`JobSchedulerTests.cs`** — 23 tests: both `AddUpdateJob` overloads, `GetAllJobs`, `RemoveJob`, dispose/post-shutdown guards, and the three job→scheduler event relays.
- **`JobQueueTests.cs`** + **`TestDoubles.cs`** — 6 tests: ctor guard, dispose idempotency, creation-failure path.

**No `Thread.Sleep` anywhere** — the fire-and-forget `Raise*` events are awaited via `TaskCompletionSource`, so tests complete the moment the event fires. No new Sonar S2925 debt.

### Measured effect (unit-tests-only basis, before → after)

| Class | before | after |
|---|---|---|
| PendingEventHeap | 0% | **100%** |
| ScheduledJob | 0% | **96.0%** |
| JobScheduler | 0% | **81.3%** |
| JobQueue | 0% | 38.0% |

898 lines newly covered by unit tests.

### Two honest caveats
1. **All four classes had zero *unit* coverage before this** — their prior 62–75% came entirely from integration tests. So these 898 lines overlap with existing integration coverage by an unknown amount, and **the net movement on the merged Codecov number will be smaller than the raw figure suggests.** The `Program.cs` denominator reduction is the one part that's independent of overlap. Worth reading the merged number on this PR rather than assuming.
2. **`JobQueue` did not reach the ~90% target.** Its queue-creation success paths need a real transport (`CreateMethodJobProducer`), which would require a `Transport.Memory` project reference that was out of scope. Integration tests already cover those paths. Guard/dispose/creation-failure branches are covered.

Verification: `dotnet build NoTests.sln -c Release -p:CI=true` → 0 warnings / 0 errors; core suite **993/993** green.

## Summary by CodeRabbit

* **Tests**
  * Added job scheduler test suites covering job queue lifecycle, validation, disposal idempotency, scheduling start/stop/update behavior, and event relay/error propagation.
  * Added coverage for pending-event heap ordering (including duplicates, resizing, and interleaved push/pop scenarios).
  * Added scheduled job tests for next-event calculation (including missed-event window logic), stale run handling, enqueue outcomes, and termination when no further occurrences exist.
* **Documentation**
  * Updated code-coverage guidance, including intentional exclusions and Codecov gate/number interpretation.


blehnen:
Addressed the nitpick: narrowed `ExcludeByFile` from `**/Program.cs` to `**/DotNetWorkQueue.Dashboard.Ui/Program.cs`.

Agreed on the substance — a bare filename glob would also match any `Program.cs` added to another project later, silently dropping real code from coverage, which is exactly the erosion this policy exists to prevent. The docs' "unambiguous" claim was true of the current tree but not enforced by the pattern; reworded.

I did not add the suggested CI guard for unexpected `Program.cs` matches. Once the pattern is scoped to a single path, a `Program.cs` elsewhere can no longer match, so the guard would be defending against a condition that cannot occur.

Verified after the change: 0 `Program.cs` entries in the coverage report, 110 other Dashboard.Ui classes still covered.

Issue #206, opened 2026-07-13
Add missing LGPL-2.1 license headers to non-test source files


blehnen:
Per the repo convention (CLAUDE.md: *"All source files include LGPL-2.1 license headers"*), several **non-test** production/library `.cs` files are missing the header. CodeRabbit flagged `Source/DotNetWorkQueue/Queue/ConsumerQueueErrorNotification.cs` during PR #205 — a scan turned up 17 in total. Not worth reopening #205 for; tracking here as a standalone cleanup.

## Files missing the header (17)

Most are the recently-added notifications feature cluster:

- `Source/DotNetWorkQueue/IConsumerMetricsNotification.cs`
- `Source/DotNetWorkQueue/IConsumerQueueErrorNotification.cs`
- `Source/DotNetWorkQueue/IConsumerQueueNotification.cs`
- `Source/DotNetWorkQueue/IWorker.cs`
- `Source/DotNetWorkQueue/Notifications/BaseNotification.cs`
- `Source/DotNetWorkQueue/Notifications/ErrorNotification.cs`
- `Source/DotNetWorkQueue/Notifications/ErrorReceiveNotification.cs`
- `Source/DotNetWorkQueue/Notifications/MessageCompleteNotification.cs`
- `Source/DotNetWorkQueue/Notifications/PoisonMessageNotification.cs`
- `Source/DotNetWorkQueue/Notifications/RollbackNotification.cs`
- `Source/DotNetWorkQueue/Queue/ConsumerMetricsNotification.cs`
- `Source/DotNetWorkQueue/Queue/ConsumerMetricsNotificationNoOp.cs`
- `Source/DotNetWorkQueue/Queue/ConsumerQueueErrorNotification.cs`
- `Source/DotNetWorkQueue/Queue/ConsumerQueueNotification.cs`
- `Source/DotNetWorkQueue/Queue/ConsumerQueueNotifications.cs`
- `Source/DotNetWorkQueue.Dashboard.Ui/Models/BulkActionResponse.cs`
- `Source/DotNetWorkQueue.Transport.SqlServer/Decorator/MockSqlException.cs`

## Explicitly excluded

- **`Source/DotNetWorkQueue/netfx/System/Guard.cs`** — do **NOT** touch. It is vendored third-party code (NETFx) that already carries its own **BSD license** header; overwriting it with LGPL would be incorrect. (It only surfaced in the scan because the grep matched the DotNetWorkQueue LGPL text specifically.)
[...]

PR #217, opened 2026-08-24
perf(sqlite): enable connection pooling by default (9.6x on single sends)

blehnen:
## Summary

Enables ADO.NET connection pooling by default in the SQLite transport. Measured single-threaded on net10: **8.54 ms → 0.89 ms per send, a 9.6× improvement**, with no change required from callers.

`System.Data.SQLite` defaults `Pooling` to `false`, unlike the SQL Server and PostgreSQL providers. The transport opens a connection per operation — the correct and idiomatic pattern for the pooled providers, but expensive here.

**The cost is not the open.** In WAL journal mode — the transport default since `EnableWalMode` — SQLite checkpoints the write-ahead log back into the database when the *last* connection to it closes. Opening and closing a connection per send therefore pays a checkpoint per message.

## How that was established

Measured against a hand-written narrow outbox table on the same engine and settings, with the journal mode verified by reading `PRAGMA journal_mode` back off the file rather than assuming the setting landed.

| config | ms/send |
|---|---|
| WAL, unpooled (previous default) | 8.54 |
| WAL, pooled (**new default**) | **0.89** |
| rollback journal, unpooled | 5.87 |
| rollback journal, pooled | 5.86 |

The last two rows are the load-bearing ones: **pooling makes no difference at all outside WAL.** That is what identifies the close-checkpoint — rather than connection-open cost — as the thing being avoided. It also means the previous default was the worst of the available configurations: WAL-without-pooling (8.54) was slower than no-WAL-without-pooling (5.87).

Two further checks that the improvement is real rather than a measurement artifact:

- The new default now matches what an explicit `Pooling=True` connection string already achieved (0.89 vs 0.90).
- The transport's fsync cost (`synchronous=FULL` minus `NORMAL`) went **0.319 → 2.506 ms**, now matching the comparison table's 2.581. The checkpoint had been swamping the fsync signal entirely; with it gone, the `FULL` control behaves as intended.

## Behavior change ⚠️

**Callers that delete SQLite database files are affected.** A pooled connection keeps the database file handle open, so `File.Delete` on a database the transport has touched will now fail where it previously succeeded.

- `SqLiteMessageQueueCreation.RemoveQueue()` calls `SQLiteConnection.ClearAllPools()` on your behalf, so the usual remove-then-delete sequence is safe.
- Callers that delete a database file **without** going through `RemoveQueue()` — test fixtures especially — must call `ClearAllPools()` first, and should delete the `-wal` and `-shm` files alongside the database.
- Opt out entirely with `Pooling=False` in the connection string. An explicit `Pooling=True` *or* `Pooling=False` is always respected.

In-memory databases are never pooled: the hold connection already keeps them alive, and with `cache=shared` a pooled connection would keep one alive past the point the caller disposed of it. That also gives `DbFactory`'s previously dead `forMemoryHold` parameter a meaning, without changing a public interface.

## Implementation note

Pooling is applied by **appending** to the caller's connection string rather than round-tripping it through `SQLiteConnectionStringBuilder`. The builder discards input it cannot parse — `"this is not a connection string=;;;"` comes back as `"pooling=True"`, which would silently replace a caller's connection string with garbage. Valid strings survive the round-trip (unknown keywords, passwords and `FullUri` were all checked), so the exposure was limited to malformed input, but there is no reason to take it. A unit test guards that exact regression.

## Changes

- `Basic/ConnectionStringPooling.cs` (new, internal) — the policy, with the result cached per connection string since it sits on the send and receive paths
- `Basic/DbFactory.cs` — `CreateConnection` routes through it
- `Basic/SqLiteMessageQueueCreation.cs` — `RemoveQueue` clears the pool after dropping tables
- `...SQLite.Integration.Tests/ConnectionString.cs` — teardown clears pools, deletes the `-wal`/`-shm`/`-journal` siblings, and no longer swallows a failed delete. With pooling on, the previous silent swallow would have leaked a database file per test across the parallel CI stages
- `...SQLite.Tests/Basic/ConnectionStringPoolingTests.cs` (new) — 12 tests: default-on, explicit true/false including casing, both in-memory forms, `forMemoryHold`, malformed input, null/empty, cache stability, and verbatim preservation of the caller's string
- `CHANGELOG.md` — entry leads with the measurement and carries the behavior warning

## Testing

Green on net10.0: **12/12** new pooling unit tests, **157/157** SQLite unit, **198/198** SQLite integration, Release build clean. The integration run left **zero** leaked database files.

## Not yet verified

All measurements are from **WSL2 ext4**. The mechanism is a SQLite checkpoint rather than a filesystem behavior, so the direction is not in question, but the 9.6× magnitude may differ on Windows/NTFS.

## Summary by CodeRabbit

- **New Features**
  - File-based SQLite databases now use connection pooling by default for improved WAL-mode performance.
  - Explicit pooling settings remain unchanged, with `Pooling=False` available as an opt-out.
  - In-memory databases continue to avoid pooling.

- **Bug Fixes**
  - Improved queue removal and database cleanup, including WAL, shared-memory, and journal files.
  - Added connection-pool cleanup, retries, and warnings when database files cannot be deleted immediately.

- **Documentation**
  - Added unreleased documentation describing pooling behavior, performance improvements, and cleanup requirements.


blehnen:
Thanks — two of the three are addressed in 0a2d4296, and I am declining the third with evidence.

## Bounded cache — accepted

Legitimate: it is a process-lifetime static, and a caller generating connection strings dynamically (one per tenant, say) could grow it without limit. Capped at 128; past the cap the value is still returned, just rebuilt each time.

I kept the cache rather than removing it because I measured what it buys: rebuilding costs **12.68 µs, about 1.4% of a send** — and that share grows as the send path gets faster, which is the point of the next phase of this work. Bounding it gets the safety without giving that up.

Added a test asserting `Apply` stays correct across 300 unique connection strings, well past the cap.

## Interlocked disposal guard — accepted

Good catch, and it matters more than it looks: `Dispose` gained a **process-wide** side effect in this PR (`SQLiteConnection.ClearAllPools()`), and these fixtures are used from parallel test runs. Guarded.

## License header on the test file — declining

This one does not match the repository’s actual convention. Headers are universal in library code and predominantly absent in tests:

| | with header | without |
|---|---|---|
| `Transport.SQLite` (library) | 96 | 0 |
| `Transport.SQLite.Tests` | 8 | **24** |
| all test projects | 264 | **731** |

The maintainer’s own license-header sweep (#206) also scoped itself to non-test files. `ConnectionStringPoolingTests.cs` matches its 24 immediate neighbours; adding a header would make it the odd one out rather than bring it into line. Happy to add it if the maintainer would rather standardise the other way, but that would be a repo-wide decision rather than something to change on this PR.

blehnen:
Second round — one accepted in db9c93a5, two declined.

## Redundant `File.Exists` guard — accepted

Correct, and slightly better than "redundant": `File.Delete` is a no-op for a file that is not there, and a missing directory surfaces as `DirectoryNotFoundException`, which derives from `IOException` and was already handled. The guard only bought a syscall per file plus a window in which the answer could change between check and delete. Removed.

## Locking cache admission — declining

The suggestion is to guard the `Count` check and `TryAdd` with a lock so admission is atomic. I do not think that trade is worth making here.

The cap exists to stop the cache growing for the life of the process, not to hold an exact count. Without a lock, the overshoot is bounded by the number of threads that happen to be inside `Apply` at the moment `Count` crosses 128 — so the worst case is a few entries over, and it is still bounded, which is the property being bought.

Against that, a lock would serialise a path that sits on **every send and every receive**. This PR exists to take a send from 8.54 ms to 0.89 ms; adding a contention point to the same path to make an approximate bound exact is the wrong direction. The existing comment says as much:

```
//Count is a snapshot, so this can overshoot slightly under concurrency. That is fine:
//the point is a bound, not an exact one.
```

If the overshoot ever mattered, the fix would be to drop the cache entirely rather than lock it — but that costs a measured 12.68 µs per connection, ~1.4% of a send today and a larger share once the send path gets faster.

## License header — declining again

Same finding as the first round; re-raised because the file changed. The evidence has not: headers are universal in library code (96/96 in `Transport.SQLite`) and predominantly absent in tests (24 of 32 without in `Transport.SQLite.Tests`, 731 of 995 across all test projects), and the maintainer’s own license-header sweep (#206) scoped itself to non-test files. Adding one here would make this file the outlier rather than bring it into line. Standardising test projects either way is a repo-wide decision, not something to settle on this PR.

PR #218, opened 2026-08-24
perf(sqlite): reuse connection objects between operations (~24x cumulative)

blehnen:
## Summary

Three commits: the connection-reuse change, the benchmark harness that found it, and a one-character Sonar fix.

**A single SQLite send goes from ~0.89 ms to ~0.36 ms** — roughly **24x cumulative** against the 8.5 ms baseline before #217.

## The finding

The expensive part of a SQLite operation is not opening the database, and it is not the provider's connection pool. **It is constructing the `SQLiteConnection` object.**

| operation | cost |
|---|---|
| `new SQLiteConnection()` — **no arguments at all** | **401 µs** |
| `new SQLiteConnection(full connection string)` | 332 µs |
| ...then `Open()` on it | +63 µs |
| assigning `ConnectionString` to an **existing** object | **0.007 µs** |

`System.Data.SQLite` pools the *native handle*, which is the cheap part — which is why `Pooling=True` from #217 recovered less than expected, why the per-open distribution was uniform rather than bimodal, and why a "pooled" open still allocated 84.7 KB.

Reusing the managed object is what matters, and **it only pays while the connection stays open**:

| strategy, one insert | cost |
|---|---|
| new connection per operation | 695.8 µs |
| object pool, `Close()` on return | 188.3 µs |
| **object pool, left open on return** | **14.8 µs** |

## Implementation

`DbFactory` hands out a `PooledConnection` wrapper for file databases, which returns its connection on `Dispose` rather than closing it.

The pool lives **inside `DbFactory`**, which is already a container singleton, so pooled connections inherit the lifetime of the queue that created them. No static state, and nothing outlives the object that created it. Bounded per connection string; connections are validated as open before reuse and disposed rather than returned otherwise; disposal is idempotent. In-memory databases and hold connections are never pooled.

`Open()` on the wrapper is a no-op when already open, because callers legitimately call it and `SQLiteConnection.Open()` would throw on an open connection. `Close()` is a no-op for the same reason the pool exists. Nothing in the transport calls `Close()`, `State` or `ChangeDatabase` — checked before relying on it.

## Behaviour change ⚠️

**Disposing the producer or consumer is what releases the database file.** Connections are held open for the queue's lifetime, so `RemoveQueue`'s `ClearAllPools()` cannot release connections held by a *live* queue in another container — they are open, not idle.

This is narrower than it sounds: queues are already `IDisposable`, and the usual `using` pattern gives the right ordering. The only caller affected is one deleting a database file **without** disposing the queue first, which was never really supportable.

## On Microsoft.Data.Sqlite

Measured as an alternative and **rejected**: it reaches 15.5 µs per insert with no pooling code, but that is the same number this change reaches on the provider already in use (14.8 µs). A migration would touch every command handler, change a public dependency, and carry dialect and type-handling differences — for no measured gain. Reported benchmarks also favour `System.Data.SQLite` for bulk inserts and mass reads. Worth noting for the future: its own pool is good enough that adding an object pool on top made it *slower*, so this code should not be ported if the provider is ever swapped for other reasons.

## Benchmark harness

`Source/DotNetWorkQueue.Benchmarks` — a BenchmarkDotNet project decomposing a send into its parts. Not shipped (`IsPackable=false`, excluded from `DotNetWorkQueueNoTests.sln`), not run by CI.

It carries a runtime guard worth knowing about: **the repository lives on `/mnt/f`, a Windows drive mounted into WSL, and .NET memory-maps assemblies — page faults against drvfs/9p impose a uniform ~35x floor on everything in the process.** Identical binaries measured 19,798 µs from `/mnt/f` and 569 µs from `/tmp`. The uniformity is what makes it dangerous: it reads as a real result. `Program.Main` now prints a red banner and the correct command when it detects this. Windows is unaffected.

## Also

Fixes CA1865 from SonarCloud on the #217 merge — `EndsWith(";")` should be `EndsWith(';')`. That was the only open issue on the project.

## Testing

- **22/22** SQLite unit tests (9 new: rent/return, reuse across operations, double dispose, use after dispose, in-memory and hold exclusions, and specifically that disposing the factory lets the database file be deleted)
- **198/198** SQLite integration tests, **zero leaked database files**, no teardown warnings
- Release build clean

## Summary by CodeRabbit

* **New Features**
  * SQLite file connections are now reused between queue operations for improved efficiency.
  * Connections remain open for the lifetime of their producer or consumer queue and are released when the queue is disposed.
  * In-memory databases and explicitly held connections continue to bypass pooling.
  * Added benchmark tools for comparing SQLite and memory transport performance.

* **Bug Fixes**
  * Improved pooled connection cleanup, disposal, transaction safety, and database-file release.

* **Documentation**
  * Documented connection reuse, file-locking behavior, disposal requirements, and benchmark usage.


blehnen:
Addressed in 6385a1c6 — 11 of 13 Sonar issues and 5 of 6 CodeRabbit findings fixed, two declined with reasons.

## Two real correctness bugs, both worth catching

**A connection disposed before its transaction was returned to the pool mid-transaction**, handing that transaction to the next renter. `SQLiteConnection.AutoCommit` is false exactly while a transaction is open — verified empirically — so such a connection is now discarded rather than pooled. Added a regression test that asserts the next renter cannot see the uncommitted row.

**`Dispose` could drain the pool between `Return`’s checks and its `Add`**, leaving a connection alive holding the database file — precisely the failure mode this PR is meant to keep under control. `Return` now adds first and re-checks disposal, which is race-free: either `Dispose` sees the connection, or `Return` sees the flag and drains.

**S4275 (CRITICAL)** was also fair: `PooledConnection.ConnectionString` returned `Inner.ConnectionString` while carrying a `_connectionString` field. Functionally equivalent, but confusing enough to deserve the flag. Returns the field now.

## Conformance

`DbFactory` is public and non-sealed, so it now implements the full dispose pattern with `GC.SuppressFinalize`, and uses `ObjectDisposedException.ThrowIf` as `ContainerWrapper` already does (S3881, CA1816, CA1513). Benchmark project smells fixed (S3241, S2486, S108 ×3, S1450 ×3). S3011 suppressed with a written justification rather than left open — reflecting into the send chain is deliberate, confined to a project that is never shipped, and measuring the *configured* serializer rather than a hand-built approximation is the entire point.

Also relabelled the two raw-table connection benchmarks as `BASELINE (provider pool)`. Good catch — they measure the provider’s pool, not the transport’s, and after this change that distinction matters.

## Declined, with reasons

**Making the pool cap atomic.** The overshoot is bounded by the number of threads returning at once, and the cap exists to *bound* the pool rather than hold an exact size. A lock there would sit on every operation — in a change whose entire purpose is removing per-operation cost. If the bound ever needed to be exact, the right answer would be a different collection, not a lock on the hot path.

**LGPL header on `DbFactoryPoolTests.cs`.** Third time this has come up, and the evidence is unchanged: headers are universal in library code (96/96 in `Transport.SQLite`) and absent from **731 of 995** test files, and the maintainer’s own header sweep (#206) scoped itself to non-test files. Accepted for the two `Benchmarks` files, which are non-test code under `Source/`.

Tests: 169/169 SQLite unit, 23/23 pooling including the new regression test. Integration re-running locally now, since `DbFactory` changed materially.

PR #220, opened 2026-08-25
perf(sqlite): cache the connection string parse, and fix per-message temp table accumulation

blehnen:
Third phase of the SQLite performance work, after #217 (pooling default) and #218 (connection object reuse). Two changes: one optimisation on the send path, and one **regression fix on the receive path that #218 introduced**.

---

## 1. Regression fix: the dequeue temp table accumulated, one per message

Each dequeue stages its candidate row in a temp table whose name came from a fresh GUID, on the stated assumption in the code that *"the temp table drops automatically when the connection closes"*. That was true until #218 made connections pooled and held open for the life of the queue. Now every **committed** dequeue leaves its table behind, for as long as the consumer runs. (An empty poll is unaffected — that transaction rolls back and takes the `CREATE` with it.)

The cost is not flat, because each `CREATE` checks a growing `sqlite_temp_master` and each one is a schema change:

| dequeues so far | create cost, before | after |
|---|---|---|
| 100 | 46.7 µs | 4.8 µs |
| 1,000 | 72.0 µs | 6.0 µs |
| 2,500 | 311.0 µs | 4.9 µs |
| 5,000 | **346.1 µs** (5,000 tables resident) | **5.2 µs** (1 table) |

…and still climbing at 5,000. **A consumer got slower the longer it ran.** This is the read-load path.

The fix derives the name from the queue's meta table instead of a GUID, uses `CREATE TEMP TABLE IF NOT EXISTS`, and clears the table before use — so after the first dequeue on a given connection the create is a no-op. Temp tables are private to a connection and live in the temp schema, so a fixed name cannot collide with a caller's own tables, and each rented connection is exclusive to one operation, so two dequeues never share one.

**No published version is affected** — #217 and #218 are still in the `Unreleased` section, so this is fixed inside the same window that introduced it.

Five tests cover it (`ReceiveMessageTempTableTests`); I verified four of them fail against the previous behaviour.

---

## 2. Optimisation: the connection string is parsed once

The harness added in #218 puts `DatabaseExists` on its own rung. On master it measured **7,238 ns and 20.2 KB per call** — a third of everything a send allocated, all of it `new SQLiteConnectionStringBuilder(...)`. A send parses **twice**: `SendMessageCommandHandler` calls `DatabaseExists.Exists` per message, and `DbFactory.CreateConnection` re-parses to decide whether the database is in memory.

`GetFileNameFromConnectionString` now caches the parse and shares the immutable `ConnectionStringInfo`, bounded at 128 entries the same way `ConnectionStringPooling` bounds its cache. Past the cap the answer is still correct, just parsed each time.

| rung | before | after |
|---|---|---|
| DatabaseExists check | 7,238 ns / 20.2 KB | **383 ns / 0 B** |
| **SQLite send, end to end** | 101,021 ns / 63.2 KB | **81,550 ns / 22.8 KB** |

−19% latency, −64% allocation; the −41.4 KB is exactly the two parses. Cumulatively across the three phases a single send has gone from ~8.5 ms (pre-pooling) to ~82 µs.

Only the *parse* is cached — `DatabaseExists.Exists` still calls `File.Exists` every time, so a database deleted underneath a live queue is still reported correctly. `AnExistingFile_IsFoundThroughDatabaseExists` deletes the file between two calls on the same connection string and asserts the second returns false. 10 tests total; there was no direct test of this class before, as `DatabaseExistsTests` mocks the parser.

---

## 3. Two attribution rungs, and one hypothesis killed

Added to the harness and recorded in its README so neither gets re-derived:

- **`INSERT … RETURNING` is not worth adopting on the single-send path.** 43,770 ns against 43,642 ns for the `SELECT last_insert_rowid()` round trip — inside the noise band. It saves 896 B and no measurable time. I expected this to be a win; it isn't.
- **Statement preparation costs roughly 2.4 µs per statement.** Reusing command objects, which is what lets `System.Data.SQLite` keep its prepared statement, took the three-statement shape from 43,642 → 36,445 ns and 5,600 → 2,432 B. That is the next lever and a larger change than anything here.

## Where the remaining send budget sits

81.5 µs against 43.6 µs for the same statement shape on a held connection — ~38 µs of library overhead, of which only ~5.8 µs is attributed (serializer 1.5, core producer pipeline 3.9, existence check 0.4). Preparation accounts for roughly 7 µs of the rest. The metadata `INSERT` is also rebuilt from a `StringBuilder` on every send while the body insert is served from `IDbCommandStringCache`.

## Verification

- SQLite unit tests: 184/184.
- SQLite integration suite: 198/198 on the parse-cache commit; re-running on the temp-table commit since it changes the receive path.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved SQLite transport performance by caching connection-string parsing results.
  * Reused temporary dequeue tables to reduce table-creation overhead and prevent table accumulation on pooled connections.
  * Added benchmark coverage for generated-ID retrieval and command reuse.

* **Reliability**
  * Improved connection pooling behavior under concurrent activity.

* **Tests**
  * Expanded validation for connection-string parsing, caching, in-memory databases, file detection, and temporary-table reuse.


blehnen:
Went through the CodeRabbit review — 2 accepted, 3 declined, reasoning below.

**Accepted**

1. **`Synchronous=Off` in the harness README was wrong.** `SendPathBenchmarks` sets `Sync = "NORMAL"`, so anyone reproducing the numbers from the README would have measured something else. My error; corrected.

2. **The 128-entry cap is now an actual bound.** Admission to the parse cache is serialised, so concurrent misses can no longer push it past the cap. This runs only when the lookup missed — after warm-up, never — so it costs nothing on the path the cache exists to protect. Applied to `ConnectionStringPooling` too, so the two caches stay identical.

   Worth noting this is deliberately *not* the call I made for the connection pool cap on #218. There the equivalent check runs on every `Return`, so a lock would sit on the hot path; here it sits on a miss. Different situation, different answer.

**Declined**

3. **Interlocked disposal in the benchmark's `[GlobalCleanup]`.** That pattern exists to guard library disposal a caller can race. `GlobalCleanup` runs once, single-threaded, in a harness that is never shipped and not in `NoTests.sln`, and the pre-existing lines beside the ones flagged dispose plainly. Adding it would make the method look thread-safety-critical when nothing about it is.

4 & 5. **LGPL headers on the two new test files.** 731 of the 995 test files in this repo carry no header, none of the sibling files in this directory do (`ConnectionStringPoolingTests`, `DbFactoryPoolTests`, `DatabaseExistsTests`), and the header sweep tracked in #206 was explicitly scoped to non-test files. Adding headers to two files would make the directory inconsistent rather than consistent. Happy to be overruled — if headers are wanted on tests, a sweep is the way to do it, not two files at a time.

PR #221, opened 2026-08-25
perf(sqlite): reuse dequeue commands so statements compile once (4.6x on the read path)

blehnen:
Fourth phase of the SQLite performance work, after #217 (pooling default), #218 (connection object reuse) and #220 (connection string parse + temp table regression). This one is on the **read** path.

---

## 1. The harness only measured sends

That is why the temp-table accumulation fixed in #220 went unnoticed — it was on the read path and only showed up over time. `ReceivePathBenchmarks` now decomposes a dequeue the way `SendPathBenchmarks` decomposes a send.

Every rung runs against an **empty** queue. A dequeue consumes the row it finds, so a populated queue would make each iteration depend on what the last one left behind, which BenchmarkDotNet cannot control. An empty poll runs the whole script — temp table, candidate select, join — and simply finds nothing, so it measures everything except materialising a row and the follow-up status commands. It is also a real workload in its own right: an idle consumer does exactly this on every poll.

`SQLiteCommand.Prepare()` is a no-op in System.Data.SQLite — statements compile lazily on execution — so compilation cannot be timed on its own. The fresh-versus-reused pair measures it where it actually happens.

## 2. Statement compilation dominates a dequeue

| rung | mean | allocated |
|---|---|---|
| generate the dequeue SQL | 526 ns | 6,848 B |
| dequeue, fresh command (**before**) | 27,760 ns | 22,144 B |
| **dequeue, through the shipped plumbing (after)** | **6,021 ns** | **7,496 B** |
| dequeue, command reused in isolation (ceiling) | 4,445 ns | 552 B |
| dequeue, command reused with parameters kept | 4,230 ns | 312 B |

**4.6x faster, 3x less allocated.** Only 526 ns of the saving is building the SQL text; the rest is recompiling the script.

Compiled statements belong to a connection, so they have to live and die with one. The pool now holds a connection **together with the commands compiled on it** (`PooledConnectionEntry`), and `PooledCommand` hands a caller the command for a given text, keeping the compiled form on dispose and clearing only the parameters.

## 3. Design choices, and one I reversed

**Callers do not change.** A cached command must hand back an empty parameter collection, because callers build their own through `CreateParameter`. I measured whether that gives the win away before building anything: 4,445 ns rebuilding parameters against 4,230 ns keeping them — 99% of it — so only the compiled statement is kept.

**`CommandText` assignment.** Setting it to the value it already holds is ignored, because the setter discards the compiled statements. Setting it to a *different* value is refused with `NotSupportedException` rather than silently leaving the command filed under a key that no longer describes it.

**The script itself is still built per dequeue, deliberately.** I first cached it per route set, then found that `GetUserClause` and `GetUserParameters` can be backed by a **caller-supplied factory that is meant to be consulted on every dequeue** — `SetUserParametersAndClause`. Caching would have silently broken that documented feature. Reverted. It leaves 6,848 B on the table, which is most of the remaining gap to the ceiling above, and it keeps the feature correct.

**Capped at 16 commands per connection.** The dequeue script embeds the caller's user clause, which a caller could vary per call. Past the cap, commands are created and disposed exactly as before — a wrong cap costs a missed optimisation, never correctness (`BeyondTheCap_CommandsStillWork`).

## 4. Public API

`IDbFactory` gains `CreateCommand(IDbConnection, string)` as a **default interface method** whose default creates a command and sets its text. Existing implementations keep compiling and working unchanged.

## 5. Also

Drops the CHANGELOG entry for the dequeue temp table from #220. It described a regression introduced and fixed inside the same unreleased window, so the net change for a user is nothing and it was treating the changelog as a history rather than a release note.

## Verification

- SQLite unit tests: **184/184**, including 12 new in `PooledCommandTests` covering reuse across pool round-trips, parameter isolation between callers, a command already in use not being handed out twice, the cap, text-change refusal, unpooled/in-memory fallback, and that disposing the factory still releases the database file with commands cached.
- SQLite integration: **198/198**, zero leaked database files.

Cumulative across the four phases: a send has gone from ~8.5 ms to ~82 µs, and a dequeue from ~27.8 µs to ~6.0 µs.

## Summary by CodeRabbit

- **Performance**
  - Improved SQLite dequeue performance by reusing prepared commands and compiled statements.
  - Reduced overhead when processing messages from SQLite-backed queues.

- **Reliability**
  - Improved connection and command pooling behavior, including safe reuse, concurrent access handling, cache limits, and cleanup during disposal.
  - Ensured pooled commands do not retain parameters or state between operations.

- **Documentation**
  - Added SQLite dequeue benchmark results and clarified performance measurement details in the changelog and benchmark documentation.


blehnen:
Worked through both reviews — 8 accepted, 1 declined. Two of them were real defects in my code.

**Real bugs, both caught by review**

1. **Cached commands leaked their settings to the next caller.** `PooledCommand.Dispose()` cleared parameters and the transaction but not `CommandTimeout`, `UpdatedRowSource` or `CommandType`. A caller that changed any of those would have passed them to whoever next asked for that SQL. The values are now recorded when the command is created — before any caller has touched it — and restored in `PooledConnectionEntry.Release`. I confirmed the new test fails without the restore.

2. **The transport-plumbing benchmark overclaimed.** I described it as "what a consumer runs". It isn't: `ReceiveMessageQueryHandler.Handle` also opens a transaction and goes through `BuildDequeueCommand`, and the rung does neither. Renamed to *"dequeue an empty queue, command from the factory cache"* and documented as command acquisition and execution only, with the note that it and the fresh-command rung omit the same things — so the **ratio** is the meaningful result, not the absolute. The README table gains the row it was missing.

   The 4.6x headline is unaffected, since both rungs were always measured the same way. But "through the shipped plumbing" in the PR description was the wrong phrase and I've stopped using it.

**Weak test**

3. `ACommandAlreadyInUse_IsNotHandedOutTwice` executed its two commands sequentially, so it passed whether or not the pool handed out the same object twice. It now marks the held command with a parameter and asserts the second one doesn't see it. (CodeRabbit suggested asserting `CachedCommandCount == 2`; that would have been wrong — the in-use path deliberately returns an *uncached* command, so the count stays 1.)

**Mechanical**

4. **S4275 ×2** — `CommandText` read through the wrapped command rather than the field it's filed under. They're equal by construction, so the property now uses `_commandText`, which is also what the getter means. Same fix as `PooledConnection.ConnectionString` on #218.
5. **S108 ×2** — the empty catch blocks in the benchmark cleanup now say why they're empty.
6. **MSTEST0037** — `Assert.IsEmpty` for the parameter assertions.

**Declined**

7. **LGPL header on the new test file.** Same as on #218 and #220: 731 of 995 test files carry none, no sibling in this directory does, and the #206 sweep was scoped to non-test files. Happy to be overruled by a sweep.

Head is `ff0d046a`. Release build clean, 13/13 command-cache tests, benchmark re-run unchanged (28,276 → 6,213 ns). Re-running the integration suite now since the release path changed.

PR #223, opened 2026-08-26
perf(sqlite): reuse send commands so statements compile once (14%)

blehnen:
Fifth phase of the SQLite performance work. #221 applied command reuse to the dequeue path; this applies the same change to the send.

## The change

The body insert, the generated-id lookup, the meta data insert and the status insert are now asked of the factory rather than of the connection, so a pooled connection hands back the statements it already compiled for them. All four handlers — single and batch, sync and async.

A command is filed under its text, so the text has to be produced before the command is asked for. `BuildMetaCommand` and `BuildStatusCommand` each split into a text builder and a parameter binder; nothing about the SQL or the parameters they produce changes. The id lookup moves to a constant for the same reason — building that string per send would have defeated the reuse.

## Measured (net10, WSL2/ext4, ShortRun, `Synchronous=NORMAL`)

| | before | after |
|---|---|---|
| **DotNetWorkQueue SQLite send, end to end** | 93,441 ns / 23,370 B | **80,213 ns / 20,674 B** |

**14% faster, 12% less allocated.** Within the same run the raw floor moved 32,114 → 31,509 ns and the DNWQ statement shape drifted the *other* way, 48,807 → 50,339 ns, so the gain is well outside the run-to-run band. (This machine was noisier than the runs behind #220/#221 — compare within a run, not across.)

**Batch sends gain most and are not in that number.** A chunk previously built and recompiled the meta data and status inserts once per message; each is now compiled once for the whole batch. The harness only measures single sends, so that is asserted from the mechanism rather than measured — worth a batch rung if we want it pinned.

## A bug I introduced, and what caught it

The first commit here broke **every async send**. `ReaderAsync` casts straight to the provider type:

```csharp
var sqlCommand = (SQLiteCommand)command;
```

and a pooled connection hands out a `PooledCommand` wrapper, so that threw `InvalidCastException`. The integration suite caught it: **40 failures**, all `SimpleProducerAsync`, `SimpleProducerAsyncBatch`, or async consumer tests seeded by those. #221 did not hit it because the receive path executes its command directly rather than through `ReaderAsync`.

`ReaderAsync` now reaches the provider command through the wrapper and rejects anything else with a clear message. I checked the whole transport — it was the only place casting `IDbCommand` to a provider type.

Second commit is that fix, kept separate so the defect and its cause stay legible in history.

## Limitation, documented and tested

Reuse depends on the statement text being identical between sends. It is — unless `AdditionalColumnsOnMetaData` is on **and** the caller varies its column names per message, since those names appear in the insert. That case misses the cache and, past its 16-command bound, resumes the previous behaviour exactly. Two tests pin the stable case, one documents the varying case deliberately.

## Verification

- SQLite unit tests: **210/210**, including 5 for statement-text stability and 3 for the async path. I verified the async guard fails against the original cast (`Unable to cast 'PooledCommand' to 'SQLiteCommand'`).
- SQLite integration: **198/198**, zero leaked database files.

Cumulative across the five phases: a send has gone from ~8.5 ms to ~80 µs, and a dequeue from ~27.8 µs to ~6.0 µs.

## Summary by CodeRabbit

* **Performance**
  * Improved SQLite message-send performance by approximately 14%, with 12% fewer memory allocations.
  * Reused prepared database operations where applicable, including metadata, status, and identifier handling.

* **Reliability**
  * Improved asynchronous SQLite command support for pooled and standard provider commands.
  * Added clearer validation when an incompatible database command is supplied.

* **Compatibility**
  * Preserved previous behavior when metadata column definitions vary between messages.
  * Reuse remains available when only message values change; differing column names continue to use the previous behavior.


blehnen:
Both Sonar issues fixed, and CodeRabbit caught a real documentation error — one that mattered more than it first looks.

**Sonar — `S1172` ×2, accepted**

`CreateStatusRecord` took an `IMessage` it no longer uses. The status insert never needed it; splitting the builder into text and parameters is what made that visible. Removed from both handlers.

**CodeRabbit — the reuse limitation was documented backwards, accepted**

I wrote that the statement text only varies when `AdditionalColumnsOnMetaData` is **on**. That is wrong in a way that would have misled a reader: caller-defined columns are named on the meta data insert when that option is on, and on the **status** insert when it is off — so the *default* configuration varies too, which is the opposite of what my note implied.

Corrected in the CHANGELOG, and a test now pins each direction: with the option off, the status text follows the caller's columns while the meta data text stays put. Varying only the column *values* changes nothing, which is the ordinary case and was already covered.

My original `TheStatusInsertIsTheSameTextForEverySend` used empty metadata, so it could never have caught this — worth noting, since it looked like coverage.

**Declined — LGPL headers on the two new test files**

Same as #218, #220 and #221: 731 of 995 test files carry none, no sibling in that directory does, and the sweep in #206 was scoped to non-test files. Happy to be overruled by a sweep rather than file-by-file.

Head is `17c75122`. Release build clean, **211/211** SQLite unit tests. No production behaviour change in this commit beyond dropping an unused parameter.

PR #225, opened 2026-08-26
perf(sqlite): cache the dequeue SQL, cutting a dequeue's allocation by 91%

blehnen:
Sixth phase of the SQLite performance work. This is the largest remaining item on either path, and it is almost entirely allocation.

## The finding

Generating the dequeue script was **6,848 B of the 7,496 B a dequeue allocated — 91% of it** — and it is pure garbage on every poll, for the life of a consumer. It is now built on first use for a given set of routes and caller clause, and kept.

| rung (empty queue, net10, ShortRun) | time | allocated |
|---|---|---|
| fresh command, script rebuilt (the original) | 29,854 ns | 22,144 B |
| command cached, script rebuilt (before this) | 7,170 ns | 7,496 B |
| **script and command both cached (now)** | **5,800 ns** | **648 B** |

**−91% allocation, −19% time** against the previous shape; **5.1× and 34×** against the shape before any of the command work.

## Why this is safe, having been backed out of #221

#221 declined to cache this because `GetUserClause` and `GetUserParameters` may be backed by a caller-supplied factory that is meant to run on every dequeue. Caching turns out to be safe regardless, because the script does not depend on what those factories return in the way that mattered:

- **Parameter values are bound, not written into the SQL**, so they cannot affect the script at all.
- **The clause *is* written into the SQL**, so it is part of the cache key. A clause that changes simply produces a different key.

Both are still read on every dequeue. `ReceiveMessage.GetDeQueueCommand` now takes the clause rather than reaching into the configuration, which makes it a pure function of its inputs — and makes the cache key checkable, which is what the new unit tests do.

The cache is bounded at 32 entries, because the key includes a clause a factory may vary.

## Coverage gap found and closed

The factory form of the user clause had **no integration coverage** — only the configuration extensions were unit-tested — and it is exactly what this change touches. `UserDequeueFromFactory` now runs a full dequeue through it.

Being precise about what that proves: each consumer has its own handler and therefore its own cache, so it demonstrates the factory path works end to end, **not** that the factories are re-consulted within a single consumer. Six unit tests in `DequeueScriptTests` pin the cache key instead, by pinning what the script does and does not depend on — including that the clause changes it, that routes change it, and that parameter values cannot.

## Verification

- SQLite unit tests: **217/217**.
- SQLite integration: **200/200** (198 plus the two new factory rows), zero leaked database files.

Cumulative on the read path: a dequeue has gone from ~29.9 µs / 22.1 KB to ~5.8 µs / 648 B.

## Summary by CodeRabbit

- **Performance**
  - Improved SQLite message receiving by reusing generated dequeue SQL and commands, reducing repeated work and allocations.
  - Added connection-string caching, persistent connection reuse, and default pooling for more efficient transport performance.
  - Improved handling of in-memory SQLite databases.

- **Compatibility**
  - Documented disposal requirements and behavior changes for applications that delete SQLite database files while connections remain active.

- **Documentation**
  - Added benchmark findings covering dequeue caching, batching, and measured performance improvements.


blehnen:
Rebased onto master now that #224 is in (both added a findings section to the harness README — resolved by keeping both). Four findings accepted, one declined.

**Accepted — the cache key could collide, and this one was real**

The key joined the clause and the routes with a separator, so a clause containing that separator could produce the same key as a different combination. That matters more than a normal key collision would: **the route count decides how many placeholders the SQL carries**, so a collision could hand back a script of the wrong shape and bind the wrong number of parameters.

It is length-prefixed now. `Key` is `internal` so four tests can pin it, including the exact colliding pair, and the no-clause-no-routes case still allocates nothing.

**Accepted — the cap is now an actual bound**

Admission is serialised. There is no eviction path, so an overshoot would have been permanent rather than transient. It runs only on a miss, which after warm-up is never.

**Accepted — the factory test now proves what it claimed to**

`UserDequeueFromFactory` counts factory invocations and asserts each consumer consulted both factories more than once. As written before it would have passed with the clause *and* the parameter list both cached, because each consumer has its own handler and its own values — I said as much in the PR description and then left it that way, which was the wrong call.

**Accepted — README rows**

Two rows described reduced benchmark shapes as if they were the whole transport path, and the first row still claimed the script is rebuilt per dequeue, which this PR changes. Both corrected.

**Declined — LGPL headers on the two new test files**

Same as #218, #220, #221 and #223: 731 of 995 test files carry none, no sibling in those directories does, and the #206 sweep was scoped to non-test files. A sweep would be the way to change it.

Head is `b8dcc6cc`. Release build clean, **221/221** unit tests. Re-running the integration suite on the rebased branch since the key encoding changed.

PR #237, opened 2026-08-31
feat(#236): opt-in System.Text.Json serializer, selected per message

### LINKED ISSUE #236: Pluggable serializers: ship System.Text.Json alongside Newtonsoft (Newtonsoft stays the default)
`ISerializer` is already a single DI registration (`container.Register<ISerializer, JsonSerializer>(LifeStyles.Singleton)`), so alternative serializers are an additive change rather than a replacement. This issue proposes shipping at least one alongside Newtonsoft.

**Newtonsoft stays the default.** It is the right default precisely because callers can queue *any* POCO — private setters, `dynamic` bodies, polymorphic types — and Newtonsoft handles that shape with far less ceremony than the alternatives. An opt-in serializer is for callers who control their message types and want the allocation back.

## The measurement that motivates this

From the core-path pass (#230, `CorePathBenchmarks`), for a 256-byte message:

| | time | allocated |
|---|---|---|
| `JsonConvert.SerializeObject` **alone** — string only, no byte conversion | 539 ns | **3.91 KB** |
| the full current body rung (`SerializeObject` + `UTF8.GetBytes`) | 616 ns | 4.30 KB |
| headers, same shape | 295 ns | 1.88 KB |

That 3.91 KB is a floor, not a tuning problem. Two candidate fixes were measured and **both failed**: writing straight to UTF-8 bytes allocates *more* (7.92 KB — `StreamWriter`'s buffers cost more than the string they avoid), and caching the Newtonsoft `JsonSerializer` is worth about 5%. Serialization is now the single largest remaining core-library cost per message, on every transport, and it cannot move without a different serializer.

## The hard part: there is no serializer marker on a message

This is the crux, and it should be settled before any code is written.

Interceptors already solved this problem — `StandardHeaders.MessageInterceptorGraph` records per message which interceptors were applied, which is what lets the AES/3DES migration work (register both, consumers pick the right decryptor per message, drain the backlog). See 0.9.43.

**There is no equivalent for serializers.** A message carries `MessageBodyType` but nothing saying *what wrote the body*. So switching a live queue's serializer would break every message already sitting in it, with no per-message way to recover. Options:

1. Add a serializer marker to the standard headers, mirroring `MessageInterceptorGraph`, and have the consumer select per message. Costs a header per message; gives a real migration path.
2. Declare serializer choice fixed for the lifetime of a queue's data, and document that changing it requires draining first.

Option 1 is the consistent one given the interceptor precedent, but it is not free and it is a wire change. Worth deciding deliberately.

## Other design questions

- **`MessageBody.Body` is `dynamic`.** Under Newtonsoft with `TypeNameHandling.Auto` this round-trips because `$type` is written. `System.Text.Json` serializing an `object`-typed property writes only the declared type's members by default — so the wrapper needs to serialize against `value.GetType()`, or become generic. This is the single biggest functional gap and should be prototyped before anything else.
- **Polymorphism and the security model.** `TypeNameHandling.Auto` plus `AllowListSerializationBinder`/`DenyListSerializationBinder` is the current accepted risk. STJ has no direct equivalent; polymorphism is opt-in per type via `[JsonDerivedType]`. That is arguably *safer*, but it is a behaviour difference, not a drop-in — and it is a large part of why an STJ serializer cannot be the default.
- **`IInternalSerializer` is also hot.** The headers go through `JsonSerializerInternal`, which uses `PrivateSetterContractResolver` for private setters. That is 1.88 KB per send. STJ needs `[JsonInclude]` or a `JsonTypeInfoResolver` modifier to match. Decide whether the internal serializer is in scope; it is used for queue configuration too, where the private-setter behaviour matters more.
- **No new dependency.** `System.Text.Json` is in-box on net8 and net10, so this can live in core without adding a package reference — worth confirming before choosing where it goes.
- **Interceptors are unaffected** — they operate on bytes, after serialization.
- **Source generation** (`JsonSerializerContext`) is where STJ's real wins are, but it requires callers to declare their message types at compile time. Probably a second step, if at all.

## Suggested shape

1. Prototype the `dynamic` body round trip under STJ. If that cannot be made to work cleanly, the rest is moot.
2. Measure it against the table above before building anything shippable — the whole point is the allocation, and an STJ implementation that has to box or double-serialize to handle `object` bodies may not actually win.
3. Settle the marker question (#1 above).
4. Only then: implementation, opt-in registration, docs, and integration tests across transports.

Note that features which vary by transport have bitten this project before (Redis history), so serializer round-trip coverage should be integration-tested per transport, not just unit-tested.

blehnen:
Closes #236 (the v1 scope described there — see "Deliberately out of scope" below).

## What

`SystemTextJsonSerializer`, opt-in. **Newtonsoft stays the default**, and that is the right default: callers can queue any POCO, and Newtonsoft handles more shapes without annotation. This is for callers who own their message types.

Measured in the repo's own harness (`CorePathBenchmarks`, net10, ShortRun), for a 256-byte message:

| | Newtonsoft | System.Text.Json |
|---|---|---|
| serialize | 682 ns / 4,464 B | **270 ns / 752 B** |
| deserialize | 768 ns / 4,552 B | **407 ns / 1,272 B** |

**~7 KB less garbage per message round trip, on every transport.** No new dependency — `System.Text.Json` is in-box on net8 and net10.

## The two hard parts

**Recovering the runtime type.** Every dequeue calls `BytesToMessage<MessageBody>` and reads `.Body` as `dynamic` — the consumer's own `T` never reaches the deserializer, so the type has to come out of the payload. System.Text.Json *serializes* an `object`-declared property using its runtime type but hands back a `JsonElement`, so a discriminator is written explicitly: once for the body, and again for members declared as `object`, an interface, or an abstract class. That factory claims only declared types nothing can be a direct instance of, which is what keeps it from recursing into itself.

**Nothing recorded which serializer wrote a body**, so switching one would have stranded everything already queued. Messages now carry `Queue-SerializerId` and `ISerializerResolver` picks the matching serializer to read them back. A queue can therefore hold messages written by more than one serializer, which is what makes a migration possible at all.

Two placement decisions worth reviewing:

- **Resolution happens in `RootSerializer`**, which every transport already reaches deserialization through — so **no transport is touched by this PR**.
- **The header is stamped in `AddStandardMessageHeaders`**, not inside the serializer. Every send path runs through it (single, batch, async, and the method and LINQ queues via `ProducerQueue`), so it does not depend on a transport serializing the body before the headers. They all do that today, but relying on it would be a silent failure the moment one didn't.

## The fallback, and why it is configurable

Messages predating the header fall back to `ISerializerResolver.Fallback`, defaulting to the serializer registered for the queue — exactly what happened before, where whatever was registered read everything. Existing users see no change.

It is configurable rather than hardcoded to Newtonsoft for correctness, not flexibility: **replacing `ISerializer` is already a supported extension point**, and anyone who has done that has a backlog written by *their* serializer. A hardcoded "no marker ⇒ Newtonsoft" would silently break them.

A message naming an unregistered serializer **throws** rather than guessing. Reading a body with the wrong serializer does not reliably fail — it can hand back a half-populated object — and a poison message is far easier to diagnose than silent data loss.

## Security

`SystemTextJsonSerializer` resolves types through the same `ISerializationBinder` the Newtonsoft path uses, so the configured allow or deny list governs both identically. There is a test that tampers with a payload's type name and asserts the deny list still rejects it.

The strict JSON encoder is kept deliberately, and commented as such: `UnsafeRelaxedJsonEscaping` would recover a few bytes on nested and generic type names, but message bodies are rendered by the dashboard, and relaxing HTML escaping on content that reaches a web UI is not a trade worth making for the size of a type name.

## Known limitations, both inherited from System.Text.Json

- A property declared as a **concrete** base class holding a derived instance needs `[JsonDerivedType]`. Properties declared as `object`, an interface, or an abstract class need nothing. There is a test pinning this that will fail if it ever starts working, so the documentation can be removed rather than quietly going stale.
- **Private setters are not restored** — which matches the Newtonsoft *message* serializer, since `PrivateSetterContractResolver` is wired only to the internal serializer. Not a regression; the test suite asserts Newtonsoft's behaviour on every case too, so a divergence in either direction shows up.

## Public API

Additive. `ISerializer.SerializerId` is a **default interface method**, so existing implementations are unaffected (same approach as `IDbFactory.CreateCommand` in 0.10.0). `IStandardHeaders.SerializerId` is new. `AddStandardMessageHeaders` takes an extra constructor argument — source-breaking only for code constructing it directly rather than resolving it.

## Verification

- **2,773 unit tests**, net10, all nine projects
- **Memory integration 60/60**, including three new tests: a producer and consumer deliberately disagreeing about the serializer, a full System.Text.Json round trip, and one asserting the marker actually survives the transport rather than the fallback silently covering for it
- **SQLite integration 200/200**

## Deliberately out of scope

`IInternalSerializer`, which writes the headers, stays on Newtonsoft. The marker lives in the headers, so making the envelope itself pluggable would require knowing how the headers were written in order to read how they were written. The spike proved the header path works (`spike/stj-serializer`) if that is ever wanted, but it needs a separate versioning answer. It is also the smaller half — 1,880 B against ~9 KB for the body round trip.

`IExpressionSerializer` (LINQ and method queues) is a separate registration and is untouched.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in System.Text.Json serialization; Newtonsoft remains the default.
  * Messages now record the serializer used to write them, supporting mixed-serializer queues.
  * Added fallback handling for legacy messages without serializer metadata.
  * Added validation for unknown or malformed serializer metadata.
* **Documentation**
  * Added guidance on registration, migration, performance, security, and limitations.
* **Tests**
  * Added coverage for serializer selection, migration, round trips, malformed payloads, and security restrictions.
* **Benchmarks**
  * Added performance comparisons between supported serializers.


blehnen:
Thanks — three of the four were real, and chasing the major one turned up a second bug you didn't flag.

**Non-string marker (major) — fixed, and you were right about the severity.** `stamped as string` on a non-string produced `null`, which `Resolve` treats as "legacy message" and answers with the fallback. That silently reads the body with the wrong serializer, which is exactly the case the resolver exists to prevent. The header arrives deserialized with `TypeNameHandling.Auto`, so a corrupt or forged blob really can put any type under that key. Now only an absent or null marker means legacy; anything else present throws with the offending type name. Two tests added.

**Auditing the transports for the same class of problem found a second one.** The Memory transport is the only one that serializes a body when the dashboard asks for it rather than at enqueue time, because it stores live objects. So the stored marker named whatever wrote the *original* message, not the bytes it had just produced — meaning the dashboard would resolve the wrong serializer on a queue whose serializer had been changed. It now restamps the marker alongside the interceptor graph it was already refreshing. Every other transport reads stored bytes and stored headers, so those match by construction.

**BOM changes (out of scope) — correct, and my fault.** My scripted edit wrote those two files as `utf-8-sig`, adding a BOM they never had. Stripped; those files now show only the one-line constructor change.

**"256-byte message" — correct.** The fixture is a message whose *body* is a 256-byte string; the serialized form is larger because it carries the wrapper and the type name. Reworded in the benchmark README, the changelog, `docs/serializers.md` and the class's own XML docs.

---

**On per-transport integration tests**, I went with the "provide evidence" branch rather than adding five test files, because the mechanism is transport-uniform and the existing suites already exercise it:

1. Headers travel as a single opaque blob on every transport — `IInternalSerializer.ConvertToBytes(headers)` on the way in, `ConvertBytesTo<IDictionary<string, object>>` on the way out. There is no per-key handling anywhere, so no transport can single out this key.
2. `Queue-MessageInterceptorGraph` already rides in that same dictionary and is read on **every** dequeue by **every** transport to select interceptors. If headers did not survive a transport, every existing consume test in all 14 Jenkins stages would already be failing.
3. `Queue-MessageBodyType` (shipped in 0.10.0) is already a *string* header in that dictionary across all transports, so the exact shape is proven.
4. The storage columns are unbounded (`VARBINARY(MAX)` and equivalents), so the added bytes cannot truncate an existing message.
5. `SerializerMarkerTests` in the Memory integration suite asserts the marker specifically — including a test whose only job is to prove the marker survives the transport rather than the fallback quietly covering for it.

The genuinely transport-specific surface was the dashboard body handlers, and auditing those is what found the Memory bug above — so the check earned its keep even though the answer was an audit rather than new test files.

Sonar is green too: S2955 (comparing an unconstrained `T` to null — now constrained to a reference type, which also states the invariant the factory already enforces), S1905, MSTEST0052 and MSTEST0037.

Issue #234, opened 2026-08-28
Performance pass: LiteDb transport


blehnen:
The SQLite transport had a full performance pass in 0.10.0 (single send ~8.5 ms → ~80 µs; empty-queue dequeue 29.9 µs / 22.1 KB → 5.8 µs / 648 B). None of the other transports have been measured. This issue covers LiteDb.

**Nothing here has been measured yet.** This is an investigation issue: establish where the time and allocation actually go before proposing a change.

## Why LiteDb is the closest analogue to SQLite

It is the other embedded, single-file transport, so more of the SQLite findings plausibly transfer here than to any of the networked transports — connection and handle lifecycle, per-operation setup, and file-level contention are all in play. That makes it the most likely of the remaining transports to have a large win, and the most likely to hit the same traps.

## The known gap

**LiteDb has no bulk batch-send path.** From 0.9.41: SQL Server, SQLite and PostgreSQL all got a true bulk insert for `Send(List<T>)`; *"transports without a bulk-insert path (Memory and LiteDb) fall back to the previous per-message loop."*

For SQLite, batching turned out to be the single largest effect measured on the send path — 17.8 µs per message in a batch against 71.9 µs sending one at a time, roughly 4×, and larger than any of the micro-optimisations. If LiteDb's `Insert`/`InsertBulk` behaves similarly, closing this gap is likely the biggest available win and should be measured first.

Note that whole-batch atomicity is the semantic the other three adopted (any failure rolls back the entire batch, and each returned `IQueueOutputMessage` reports the failure). Whatever LiteDb does needs to match that contract, and scheduled-job messages are rejected on the batch path.

## Other things worth looking at

- **Connection mode** — shared, direct and memory modes have very different locking and handle behaviour. Measure each; the default may not be the fast one.
- **Connection and handle lifecycle** — the SQLite pass found that constructing the connection *object* (~400 µs) dwarfed opening the database (~63 µs). Check whether LiteDb has an equivalent asymmetry before assuming where the cost is.
- **Per-operation setup** — collection lookups, index checks, BSON mapper construction; anything rebuilt per message.
- **The BSON mapper** — whether it is cached per type or resolved per message.
- **Deletion and file-handle behaviour**, since the SQLite pass changed this observably for callers and it is the kind of thing that surprises people on upgrade.

## Related

[...]

PR #238, opened 2026-08-31
perf(litedb): scope the send and receive locks to a database, not the process

blehnen:
First of three for #234. This one is the safe, self-contained part; the bulk batch path and shared-mode connection reuse follow separately, since mixing a low-risk change with two that need care makes for a hard review.

## The issue guessed wrong, and measuring first is why we know

#234 assumed the missing bulk-send path would dominate. It matters — see below — but it is not the ceiling. Three **process-wide** locks were.

`SendMessageCommandHandler`, its async twin and `ReceiveMessageQueryHandler` each held a `static readonly object`. Every LiteDb queue in a process waited on every other one, **including queues writing to entirely separate database files**.

200 sends, before:

| | time | vs 1 thread |
|---|---|---|
| 1 thread | 28.7 ms | 1.00 |
| 4 threads, one queue | 38.7 ms | **1.35× slower** |
| 8 threads, one queue | 39.6 ms | 1.38× slower |
| 4 threads, **two separate queues** | 40.3 ms | **1.40× slower** |

Concurrency made it strictly worse, and two unrelated queues performed the same as one — the tell that the lock was process-wide.

## After

| | before | after |
|---|---|---|
| 4 threads, **two separate queues** | 40.3 ms | **22.4 ms** |

1.8× faster, and from *slower* than single-threaded (1.40×) to faster (0.77×).

## What changed

- Locks keyed by the database they protect rather than `static`.
- The **send** lock is taken only when a message carries a job name. That check-then-act — "is this job already queued" followed by the insert — is the thing the lock's own comment says it exists for. An ordinary send is covered by the transaction.
- The **receive** lock is narrowed but **kept**. Its comment is correct that `BeginTrans` does not block in direct or memory mode, so removing it would let two consumers claim the same record. Narrowing process-wide → per-database only removes exclusion between different databases, which never needed it.
- The key is the **resolved full path**, not the connection string, so two connections reaching the same file by different text still exclude each other. Getting that wrong on the receive path means double delivery, so it has tests.

## What this PR does not claim

Concurrency against a *single* database is unchanged (1.35× → 1.31× slower on four threads). Rather than assume that was ours to fix, the harness includes a control that does the same transaction-wrapped writes straight into LiteDB with no transport in the way:

| control | time |
|---|---|
| raw LiteDB, DNWQ-shape writes, 1 thread | 12.13 ms |
| raw LiteDB, DNWQ-shape writes, 4 threads | **20.86 ms (1.72× slower)** |

LiteDB itself scales *worse* than the transport does. A write transaction takes an exclusive engine lock, so that ceiling belongs to the storage engine. Worth not chasing in the transport later.

## Two findings recorded for the follow-ups

Both are in the benchmark README so they are not re-derived:

- **Batching is currently worse than not batching** — 207 µs per message in a batch of 100, against 149 µs sending them one at a time. With no bulk path, `SendMessages` falls back to `Parallel.ForEach` over single sends, which fans threads into that same exclusive write transaction and buys contention instead of parallelism. This is PR 2.
- **A shared connection costs 17×** — 2,506 µs against 149 µs for the same send, because `LiteDbConnectionManager.GetDatabase` constructs a new `LiteDatabase` on every operation. This is PR 3.

Also measured and deliberately *not* acted on: the per-send existence check parses the connection string every time and is not cached, but it is 0.9 µs against a 149 µs send — under 1%, and not the lever it was on SQLite.

## Verification

- **174 LiteDb unit tests**, including new coverage for the key (same file via different connection strings → same key; different files → different keys; in-memory keyed separately) and the locks (same database → same lock; different databases → different locks; jobs and de-queues do not share one)
- **LiteDb integration 89/89**, run twice — the second time after the in-memory keying fix

## Summary by CodeRabbit

* **Reliability**
  * Fixed a race condition that could duplicate synchronous and asynchronous sends for the same scheduled job.
  * Preserved safeguards against duplicate scheduled-job claims and existing dequeue behavior.

* **Serialization**
  * Continued support for opting into `System.Text.Json`, while retaining Newtonsoft.Json as the default.
  * Improved compatibility with legacy serialized messages.

* **Performance**
  * Reduced overhead for regular LiteDB message sends.
  * Improved coordination of scheduled-job processing across send methods.


blehnen:
Good catch, and it was worse than the comment says — checking it showed the change should not have been made at all.

**The symlink finding is correct, and it was a regression.** `Path.GetFullPath` does not resolve symbolic links, so a real path and a symlinked path to the same file produced different keys and therefore different de-queue locks. In direct mode that lets two consumers claim the same record. Before this branch the static lock made it impossible, so I introduced the hole. Hard links make it worse: nothing path-based resolves those at all, so canonicalizing the path would only have narrowed the window, not closed it.

**Verifying it exposed the real problem: none of that change was measured.** The concurrency benchmark only sends, and every send in it carries no job name — so it never took a lock either way. Neither the per-database keying nor the narrowed receive lock moved a single number in the PR description. The entire 1.8× came from one thing: not taking the lock when there is no job name.

So both risky halves are gone rather than repaired:

- `ReceiveMessageQueryHandler` and `LiteDbConnectionManager` are **byte-identical to master** again
- `DatabaseLocks` and `DatabaseKey` are deleted, along with their tests
- what remains is the gate — an ordinary send skips the lock, a job send takes the same process-wide lock it always did

The transport diff is now two files and one condition.

**Re-measured, and the win is intact:** 200 sends across two queues on four threads is **21.87 ms against 40.31 ms** on master, ratio 0.67 — slightly better than the 22.42 ms the more elaborate version managed.

I recorded this in the benchmark README as a finding rather than dropping it quietly, because narrowing those locks looks like the obvious next step and is a trap:

> Narrowing the locks per database was tried, measured at nothing, and reverted. The benefit comes entirely from *skipping* the lock, so keying it per database moved no number. It also cannot be done safely from a path: `Path.GetFullPath` does not resolve symbolic links, and nothing path-based resolves hard links, so two connection strings reaching one file could get different locks. On the de-queue path that means two consumers claiming the same record. Do not re-derive this without a real file-identity check.

---

Also addressed from the same review: the LiteDb section reused the `## Findings` heading that already belonged to the SQLite section (now `## LiteDb findings`), and both benchmark class comments still described the pre-fix locking — stale in the very PR that changed it.

And the earlier critical one, the project reference: fixed in `d844212`. Worth noting it built locally because WSL on a Windows drive is case-insensitive, so I audited all 111 `ProjectReference` paths in the repo against git's exact-case file list rather than fixing the one I could see — that was the only break.

Verified on the current head: Release build with `-p:CI=true` clean, 168 LiteDb unit tests, LiteDb integration 89/89.

blehnen:
Correct again, and this one is a genuine bug — but worth being precise about its age: **it predates this branch.** On master both handlers already had their own `private static readonly object Locker`:

```
master  SendMessageCommandHandler.cs:38       private static readonly object Locker = new object();
master  SendMessageCommandHandlerAsync.cs:39  private static readonly object Locker = new object();
```

So a `Send` and a `SendAsync` of the same scheduled job have never excluded each other. Both can read `NotQueued` before either commits, and both insert. My change didn't introduce it and didn't widen it — job sends behave exactly as they did — but it is real, it is in the code this PR is already rewriting, and leaving a known duplicate-insert path there would be odd. Fixed here, and called out in the changelog as pre-existing rather than dressed up as part of the perf work.

Both handlers now take one shared lock through a small `ScheduledJobLock`, which also carries the job-name condition.

## That also turned out to be the Sonar duplication

The gate failed again at 4.8%, and I had assumed it was the benchmarks. It wasn't — querying the measures by component put all 24 duplicated lines in the two send handlers:

```
SendMessageCommandHandler.cs        new_duplicated_lines=14   density=43.9
SendMessageCommandHandlerAsync.cs   new_duplicated_lines=10   density=42.9
```

It was my own `Monitor.Enter` / `finally` pair, written out once per handler. Extracting it removes the duplication and fixes the shared-lock bug in the same change. (The density had *risen* from 4.1% to 4.8% purely because deleting `DatabaseLocks` shrank the denominator.)

## Two details worth flagging for review

**`using (structExpression)` does not box.** I probed it rather than assuming: **0 bytes allocated over 1,000,000 ordinary sends.** So an ordinary send still allocates nothing to decide it needs no lock.

**The block structure is kept on purpose.** I first replaced `lock (Locker) { … }` with a `using var` declaration, which left the body over-indented; re-indenting it would have turned a method that is already near-identical between the two handlers into a large block of *new* duplicated lines and made the gate worse. Using a `using` *statement* keeps the braces, so the diff is one statement per handler and the body is untouched.

## State

The transport diff for this PR is now: one statement changed in each send handler, one new 60-line file, and the two `static` lock fields removed.

Re-measured after the rewrite — the win is unchanged: **200 sends across two queues on four threads, 21.34 ms against 40.31 ms** on master.

Verified: Release build with `-p:CI=true` clean; 174 LiteDb unit tests, including new coverage that an ordinary send is never blocked, that two scheduled jobs exclude each other, that the lock is released with its scope, and that disposing a scope which took no lock does not throw `SynchronizationLockException`; LiteDb integration 89/89.

blehnen:
Both fair, and the second one was hiding a worse problem than it reported.

## The struct scope is gone

Correct that a copied `Holder` carries `_taken == true` in every copy, so disposing two of them calls `Monitor.Exit` twice and the second throws `SynchronizationLockException`.

It is not reachable from either call site — both are `using (ScheduledJobLock.AcquireIfJob(jobName))`, and a `using` statement never copies — but that is an argument for the current code being lucky, not correct. A mutable struct implementing `IDisposable` is a footgun in a type someone will eventually assign to a local and pass around, so I removed the shape rather than documenting around it.

It is now two reference types: one **cached** no-op instance for ordinary sends, and a small class per scheduled send whose release is guarded with `Interlocked` so a second `Dispose` is a no-op.

**Measured rather than assumed**, because the obvious objection is that classes allocate:

| path | allocation over 1,000,000 iterations |
|---|---|
| ordinary send | **0 bytes** |
| scheduled send | 32 bytes each |

The hot path is unchanged because the no-op scope is a single shared instance. A scheduled send now allocates one object, on a path that already takes a process-wide lock and runs a query — a trade I am happy to make to delete a class of bug.

## The test point was right, and hardening it found a real flaw

You are right that the helper could not distinguish "blocked on the lock" from "never got scheduled" — the exclusion test would have passed whether or not the lock worked at all. Every test now signals immediately before attempting acquisition and requires that signal first.

Adding the `worker.Wait` assertion you suggested then **failed the exclusion test**, which turned out to be correct: in the blocked case the worker *cannot* finish, because the test itself is still holding the lock. My shared helper was trying to behave two different ways depending on the caller, and the assertion exposed it.

So the tests are written out individually now rather than sharing that helper. The exclusion test also proves the worker gets through **after** the lock is released, so it demonstrates exclusion rather than deadlock — which the original never did. They run in 1s instead of 4s as a side effect of using a short wait only where the assertion is that something does *not* happen.

New coverage: an ordinary send is never blocked (null/empty/blank job names), two scheduled jobs exclude each other and then proceed, disposing a scope twice is safe, disposing an ordinary send's scope twice is safe, and the lock is released with its scope.

---

Verified: Release build with `-p:CI=true` clean; **175 LiteDb unit tests**; **LiteDb integration 89/89**.

PR #239, opened 2026-09-01
perf(litedb): send a batch in one transaction instead of a parallel loop

blehnen:
Second of three for #234. PR 1 (#238) removed the process-wide lock; this one is the batch path, which is where the remaining cost lived.

## Batching was a pessimisation

LiteDb had no batch handler, so `Send(List<T>)` fell back to the loop in `SendMessages`: a `Parallel.ForEach` over single sends. Every message paid its own connection, existence check and transaction, and the threads fanned out into LiteDB's **exclusive write transaction** — so it bought contention rather than parallelism and was *slower per message* than not batching at all.

| | before | after |
|---|---|---|
| batch of 100 | 21,057 µs | **2,450 µs** (8.6×) |
| per message | 211 µs | **24.5 µs** |
| allocated | 11,203 KB | 8,624 KB (−23%) |

**Batching is now 5.8× better than sending one at a time**, against 145 µs for a single send — where before it was 1.5× worse.

## The ceiling was measured before any of this was written

Rather than assume "one transaction will help", I added raw-LiteDB rungs first:

| rung | per message |
|---|---|
| raw LiteDB, body + meta writes, one transaction | **9.3 µs** |
| raw LiteDB, `InsertBulk` | 5.0 µs |
| this PR, end to end | 24.5 µs |
| raw LiteDB, one write per transaction | 65.0 µs |

So amortising the transaction is worth 7× on its own, and the remaining ~15 µs above the ceiling is serialization and the core producer pipeline rather than storage — which matches what those cost elsewhere.

**`InsertBulk` is deliberately not used.** It is faster still, but returns a *count* rather than the generated ids, and the caller needs the ids. That is recorded in the benchmark README so it is not re-derived as an obvious missed optimisation.

## Semantics

Follows the three relational transports from 0.9.41:

- **Whole-batch atomic.** Any failure rolls back every message and each returned `IQueueOutputMessage` reports it. ⚠️ This is a behavior change: previously one bad message failed alone and the rest committed. Callers needing per-message isolation should loop `Send(message)`.
- **Ids in caller order**, since the inserts are sequential inside the one transaction.
- **Scheduled jobs rejected.** Not an oversight: the "is this job already queued" check is a check-then-act, and serving it in a batch would mean reintroducing the process-wide lock #238 just stopped taking on every send. The exception says to send them individually.

The sync and async handlers are thin wrappers over one shared implementation — LiteDB has no async API, so the async one runs it on a task exactly as the single-message handler does. That also keeps the two from drifting apart, and avoids the duplication Sonar flagged on the last PR.

## Verification

- **179 LiteDb unit tests**, including new coverage that a scheduled job anywhere in a batch is refused, that a blank job name is not treated as a job, and that an ordinary batch passes
- **LiteDb integration 93/93** — the existing `SimpleProducerBatch` / `SimpleProducerAsyncBatch` tests now exercise this path across **direct, memory and shared** connections and with the status table on and off, plus four new tests for ids being returned in caller order, unique and real, and for an empty batch
- Pre-checked both traps from the last PR: no BOM changes to existing files, and all project references verified against git's exact-case list

## Also included

The doc note agreed on #237: `docs/serializers.md` now records that the System.Text.Json **header** path was prototyped and works, so the obstacle to making `IInternalSerializer` pluggable is versioning rather than feasibility. That was the last unique content in `spike/stj-serializer`, which can now be deleted.

## Summary by CodeRabbit

- **New Features**
  - Added LiteDB batch message sending with unique IDs returned in the original order.
  - Empty batches complete successfully without sending messages.
  - Batch sends support route, delay, expiration, and status options.
- **Bug Fixes**
  - Batch failures now roll back the entire batch, with every result reporting the failure.
  - Scheduled jobs are rejected in batch sends.
- **Performance**
  - Improved 100-message batch times from roughly 21 ms to 2.45 ms.
- **Documentation**
  - Updated LiteDB batching performance findings and serializer documentation.


blehnen:
Fixed in `81690634` — and it's a fair catch, because the test was genuinely vacuous.

The batch handler returns an empty result when the database doesn't exist (mirroring what the single-send handler does). So if `CreateQueue()` had failed, `Assert.IsEmpty(results)` would have passed for that reason and the empty-batch path would never have been exercised at all. It now asserts `created.Success` with the error message, matching the other test in the file.

This is the same shape as the flaw you found in the lock tests on #238 — an assertion that can't distinguish the behaviour under test from the setup not having happened. Worth me watching for as a pattern rather than fixing case by case.

I checked the one analogous line elsewhere, in `SerializerMarkerTests` (Memory transport, merged in #237). It's **not** affected: those tests assert a message is consumed with the expected body and a specific marker header, so a missing queue fails them outright rather than passing silently. Left alone rather than dragging an unrelated transport into this PR.

Verified: 4/4 on `BatchSendResults`.

PR #241, opened 2026-09-01
perf(litedb): stop scanning the queue to find the next message

blehnen:
The LiteDb receive path, following #238/#239/#240. This one changes a query, not a lock.

## The de-queue scanned the queue

The eligibility tests were `Where` clauses. LiteDB uses **one index per query**, and it chose `Status` — where every waiting message holds the same value — so it selected the entire backlog and then sorted it. A de-queue therefore got slower the deeper the queue was.

Measured on the query that finds the message:

| waiting messages | before | after |
|---|---|---|
| 1,000 | 2.3 ms | **55 µs** |
| 10,000 | 31 ms | **55 µs** |
| 10,000, allocation | 22 MB | **157 KB** |

It now walks the collection in primary-key order and applies the tests over a window of 64. Constant with depth, and when the whole head of the queue is deferred it is still **2.5× better** than the scan it replaces.

*(These are query-level numbers. An end-to-end de-queue adds fixed costs on top — deserialization, the status update — so the whole operation improves by less than the ratio above.)*

## Keyed on `Id`, not `QueuedDateTime`

Two reasons, and the first is a correctness one. `Id` is unique, so paging cannot step over messages that share a value — a batch stamps many messages the same millisecond, and seeking by timestamp would silently lose them. And it is the primary key, so it is always indexed: **no schema change, no migration**, existing queues get this on upgrade.

## Three things measurement overturned

All recorded in the benchmark README so nobody re-derives them:

- **Adding an index for the sort field does nothing** (1.04×). The planner keeps choosing `Status`. An earlier revision of that README claimed a 931× win here — it had been measured against a collection with *no* indexes, which is not what ships. `MetaDataTable` builds four.
- **`Skip` paging is not equivalent to seek paging.** Identical when the head of the queue is ready; when it is deferred, `Skip` costs **375 ms and 1.5 GB** — twelve times worse than the scan — because it re-walks from the start on every page.
- **Dropping the `Status` and `HeartBeat` indexes is faster still, and costs more than it saves.** 25 µs instead of 55, but the heartbeat monitor is the one query where `Status == Processing` genuinely is selective, and it goes **26× slower**. Keeping every index and changing the query gets nearly all the win with no regression anywhere.

## ⚠️ Behaviour change

A message **requeued from the dashboard now goes to the back of the queue**, rather than to the position its original enqueue time would have given it. Ordering is by insertion and a requeue inserts a new record. The original enqueue time is still stored and still shown in the dashboard; both requeue sites carry a comment saying it no longer decides position.

Preserving the old behaviour would mean ordering by timestamp, which brings back the tie bug, needs a new index, and costs ~16% on inserts.

## One guard test worth looking at

`DateKindIsPreserved` pins the assumption the comparison rests on. The stored times come back as UTC, so they are compared with `UtcNow` directly — but LiteDB returns a plain `DateTime` as `Local` in other shapes, and comparing one of those with `UtcNow` compares raw ticks without applying the offset. That releases delayed messages early **while every other test still passes**. It happened to an earlier version of the benchmark in this branch, which is why the guard exists rather than a comment.

## Verification

- **179 LiteDb unit tests**
- **LiteDb integration 99/99**, including four new tests: FIFO order across 25 messages, a delayed message not consumed before it is due, an expired message not consumed, and the date-kind guard. The delay and expiry tests each send a control message alongside, so "nothing arrived" cannot pass by the consumer simply being slow

Production change is one file: `ReceiveMessageQueryHandler.cs`. No other transport is affected — none of them compares a stored date in .NET; they filter in SQL.

## Summary by CodeRabbit

- **Bug Fixes**
  - LiteDB polling now checks a bounded set of messages and resumes efficiently, improving performance for large backlogs and queues with no eligible matches.
  - Delayed and expired messages remain excluded until eligible.
  - Requeued messages are placed at the back of the queue while retaining their original enqueue time for dashboard display.

- **Tests**
  - Added coverage for FIFO processing, eligibility, deferred timestamps, and continuation across poll boundaries.

- **Documentation**
  - Updated changelog and benchmark documentation with queue performance findings.


blehnen:
Both correct, and the first one was a real regression rather than an overstatement.

## The unbounded no-match path

You were right that `PageSize` bounds one query and not the total work. I measured it rather than reasoning about it, and it is worse than "restores queue-depth-dependent I/O" — it was **worse than what it replaced**:

| nothing eligible, 10,000 rows | |
|---|---|
| as shipped | 7,122 µs |
| my walk (unbounded) | **12,051 µs — 1.7× worse** |

Materialising rows into .NET costs more than letting the engine reject them, so the walk only wins while it can stop early. On the path an idle consumer takes most often it could not, and the PR description claimed "constant with depth", which was wrong for that path.

**Fixed by bounding and resuming.** A poll reads at most sixteen pages and remembers where it stopped; the next one carries on from there:

| nothing eligible, 10,000 rows | |
|---|---|
| as shipped | 7,122 µs |
| bounded walk | **1,171 µs — 6× better** |

Nothing starves: a fruitless poll advances the position, reaching the end resets it to the head, and a message that becomes eligible behind the position is found on the next pass. Finding a message also resets it, so consecutive de-queues stay in order rather than continuing from wherever the last search ended.

**`A_Message_Beyond_One_Polls_Reach_Is_Still_Found`** puts the only ready message past what a single poll examines — 1,200 deferred messages ahead of it — so it can pass only if the resume works. I disabled the resume and watched it fail before keeping it; nothing else in the suite comes near that boundary.

I also tried your other suggestion — keeping the predicates in the engine and ordering by `Id` so the planner walks the key index and stops early. It does not work: **941 µs against 55 µs** for the walk, because the planner still selects and sorts. Recorded in the benchmark README so it is not re-derived.

## The requeue note

Also correct, and I checked the code rather than taking it on faith. `DashboardRequeueAllErrorMessagesCommandHandler` has both paths — `meta.Update(existing)` at line 78 when a meta record survives for that `QueueId`, which keeps its identifier and therefore its place, and `meta.Insert(newMeta)` at line 96 when none does. The single-message handler only inserts.

The changelog now says exactly that: the insert paths move a message to the back, the update path leaves it where it was.

---

Verified: Release build with `-p:CI=true` clean; 179 LiteDb unit tests; LiteDb integration **100/100**.

blehnen:
Right again, and this one mattered more than "the rung measures the wrong thing".

Resetting the position every invocation left the **load-bearing assumption unmeasured**: that seeking to a later position costs no more than seeking to the start. `Skip` looked fine on exactly that reasoning and turned out to be quadratic, so applying it a second time without checking would have been careless.

**It holds.** Carrying the position so the mean covers a whole sweep rather than its first pages:

| nothing eligible | shipped | unbounded walk | bounded + resuming |
|---|---|---|---|
| 1,000 rows | 926 µs | 1,195 µs | **605 µs** |
| 10,000 rows | 7,082 µs | 12,390 µs | **1,239 µs** |

1,239 µs against the 1,171 µs the first-pages rung reported — flat, where `Skip` would have climbed. Seeking an index to position 8,000 costs what seeking to 0 costs.

It also corrected a number in the *other* direction, which I would have shipped wrong. At a thousand rows the resuming poll is **605 µs against 926 µs** for the original query — not the 1.2 ms the non-resuming rung showed. So the bounded walk beats what it replaces at **every depth measured**, not only at ten thousand. The changelog and README said otherwise and now don't.

Benchmark and documentation only; no production code changed this round, so the 100/100 integration run on `0e3619f7` still stands.

Issue #235, opened 2026-08-28
Performance pass: Memory transport (and remaining core-path overhead)


blehnen:
The SQLite transport had a full performance pass in 0.10.0. None of the other transports have been measured. This issue covers the Memory transport.

**Lowest priority of the sweep**, and worth being explicit about why it is here at all.

## Why it matters despite being in-process

The Memory transport stores the POCO directly — no serialization, no SQL, no I/O. That makes it the cleanest available measurement of **core library overhead**, which is why `SendPathBenchmarks` already carries it as the *"core producer pipeline (Memory transport)"* rung. Whatever that rung costs cannot be fixed in any transport, so it doubles as the regression guard for core-path work.

Current state of that rung, after #230: **2,902 ns and 3,088 B per send** (was 4,005 ns / 4,512 B).

So a pass here is really two questions:

1. **What is left in that 3,088 B?** #230 removed the Guard expression trees (~770 B here) and the per-send `Assembly.GetName()` (520 B). The remaining ~3 KB is message and header construction, `QueueOutputMessage`, the correlation id, the dictionaries, and the decorator stack. Attributing it would tell us whether there is another core-library win of the same size, and any such win moves *every* transport.
2. **Is the Memory transport itself doing anything silly?** It has no bulk batch path (falls back to a per-message loop, per 0.9.41), and `Send(List<...>)` uses `Parallel.ForEach` into a `ConcurrentBag`, which is a curious shape for an in-memory store.

## Things worth looking at

- Attribute the remaining per-send allocation in the core pipeline rung — new rungs for header generation, message factory, and the decorator stack.
- **The decorator stack** (policy, history, metrics, trace) measured on its own. Note the Polly send decorator was already checked: nothing registers a `SendMessage` pipeline by default, so its closure never allocates.
- The `Parallel.ForEach` batch shape — likely slower than a straight loop for an in-memory store, and it was the source of the ordering bug fixed in 0.9.41.
- Whether the consume path has a similar core-overhead story; roughly 19 argument-validation calls per message consumed were measured before #230, so the consume path was never decomposed the way the send path was.

## Not in scope

Serialization. The core serializer costs ~3.9 KB per 256-byte message and that is a Newtonsoft floor, not a tuning problem — `SerializeObject` alone, producing only a string, is 3.91 KB. It does not appear in the Memory rung at all, and moving it is a separate design decision.

## Method

[...]

PR #242, opened 2026-09-02
perf(core): cut what every send allocates, and fix the Memory batch ordering (#235)

blehnen:
Closes part of #235 — the send half. The consume path is still open (see *What is left* below).

## What this changes

The Memory transport stores the POCO in a dictionary — no serialization, no SQL, no I/O — so whatever a send costs there is core library, and a change to it moves **every** transport. `SendPathBenchmarks` already carried that as one rung; this PR takes the rung apart (`MemoryPathBenchmarks`) and acts on what it found.

### Where the 3,219 B of a core send actually went

| part | bytes | share |
|---|---|---|
| `new AdditionalMessageData()` | **1,832** | **57%** |
| the Memory store write | ~700 | 22% |
| `IMessageFactory.Create` | 248 | 8% |
| policy/history/metrics decorators | ~206 | 6% |
| `GenerateMessageHeaders.HeaderSetup` | 112 | 3% |
| `AddStandardMessageHeaders.AddHeaders` | 24 | 1% |

One `AdditionalMessageData` is constructed for every message sent — `ProducerQueue.Send` makes one when the caller supplies no data — and its constructor eagerly created four collections that most messages never put anything in. The largest was a `ConcurrentDictionary`, whose lock array is sized from the processor count.

Fixed, each measured:

- **`AdditionalMessageData`** — collections created on first use. Measured as a pair in one run: **1,832 B → 72 B**, and roughly an order of magnitude less time.
- **`Message`** — a second dictionary for internal headers that most sends never touch, also created on first use. `IMessageFactory.Create` 248 B → 168 B.
- **`IAdditionalMessageData.Headers`** — built a fresh `ReadOnlyDictionary` on every read, and `HeaderSetup` read it twice. The property now hands back one cached view that reads the dictionary rather than wrapping it — which keeps the old behaviour of showing a header set after the view was taken, now that the dictionary itself may not exist yet — and `HeaderSetup` reads it once instead of twice: 112 B → 72 B, which is now just the correlation id.
- **The metrics decorators** — `ITimer.NewContext()` built a context object holding a `Stopwatch` (both classes) for a measurement the histogram then discarded when no collector was subscribed. It now returns a shared do-nothing scope when the instrument has no listener, and keeps the start as a raw timestamp when it does: the send decorator stack ~206 B → ~41 B.

### Results

Same rung, same harness, before and after:

| | before | after |
|---|---|---|
| core producer pipeline (Memory transport) | 2,990 ns / 3.02 KB | **2,190 ns / 1.15 KB** |
| SQLite send, end to end | 17.64 KB | **15.62 KB** |
| SQLite 100-message batch | 1,562 KB | **1,364 KB** |

The SQLite rows were measured back to back on the same machine from two build outputs, not against a number recorded in an earlier session.

## ⚠️ Behaviour fix: Memory batch results came back out of order

`Send(List<T>)` fanned out with `Parallel.ForEach` into a `ConcurrentBag`, so results came back in whatever order the threads finished. `SendAsync(List<T>)` filled the same kind of bag from a sequential loop — the order is lost just as thoroughly and less visibly.

0.9.41 declared batch results to be in caller input order and every other transport returns them that way, so **a caller matching a generated id back to the message it sent by position was getting the wrong message**. Both are now ordered loops.

Nothing is given up for it: the store behind this transport is an in-memory concurrent dictionary, so a send waits on nothing the fan-out could overlap, and the two shapes measure the same speed to within the noise.

## Tests

- 9 new unit tests (`AdditionalMessageDataTests`, `TimerNetTests`) covering the lazy collections, the header view held across a later write, `TryGetSetting` with no settings, and both halves of the timer's listener check.
- 4 new integration tests (`BatchSendOrder`) for sync and async batch ordering at 100 and 1,000 messages.
- Every new test was checked against the old code: the ordering tests fail 4/4 on the previous shapes, and the timer test fails when the listener check is removed. None of them are vacuous.
- All 9 unit-test projects green (2,482 tests). Memory integration suite green (64). `NoTests.sln -c Release -p:CI=true` clean.

## What is left on #235

The consume path was not decomposed — roughly 19 argument-validation calls per message consumed were counted before #230, but the receive side has never had the ladder treatment the send side has. That is worth its own pass. The issue stays open.

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory transport batch sends now return results in the same order as submitted messages for synchronous and asynchronous operations.

* **Performance**
  * Reduced allocations during message creation, header access, and metrics timing.
  * Improved efficiency when metrics collection is inactive.
  * Streamlined header processing and message send operations.

* **Documentation**
  * Updated performance measurements and added benchmarks covering Memory transport, batching, allocations, and timing behavior.


blehnen:
All three review points addressed in 64d2264a.

**`Headers` live-view behaviour** — you were right, and I had talked myself out of it on the grounds that `GenerateMessageHeaders.HeaderSetup` is the only reader of this property in the repo. That is true but irrelevant: it is a public property on a public class, and the previous contract was that a caller holding the returned view saw headers set afterwards.

Fixed without giving the allocation back: instead of wrapping the dictionary once (which cannot work when the dictionary may not exist yet), `Headers` returns a single cached `HeaderView` that reads the field on each access. Old behaviour preserved, dictionary still uncreated for the messages that carry no user headers. It costs 24 B per message, so the producer pipeline is **1.15 KB** rather than the 1.13 KB the first commit measured — the changelog, the benchmark README and this PR body are updated to the corrected numbers rather than left at the flattering ones.

New test `Headers_Held_From_Before_A_Write_Still_Shows_It` pins it, and exercises the whole view surface (indexer, `Keys`, `Values`, `Count`, `ContainsKey`, `TryGetValue`, enumeration). It fails if the view is frozen — verified.

**Benchmark comments** — both stale descriptions fixed: `BatchShape_Parallel` now says it is the shape the transport used *before* this work and is kept so the answer stays checkable, and the `HeaderSetup` comment says one read rather than two. Added a caveat there too: that rung reuses one data object, so the view's own 24 B is paid per message in the end-to-end rung rather than in the component rung.

**License header** — added. Test projects in this repo are mostly without one (4/22 in this project), but CLAUDE.md asks for it on all source files, so the instruction wins over the legacy files.

PR #244, opened 2026-09-02
perf(core): take the consume path out of the DI container (#235)

### LINKED ISSUE #235: Performance pass: Memory transport (and remaining core-path overhead)

blehnen:
The consume half of #235, and what closes it. The send half shipped in #242.

## What the ladder found

`MemoryReceiveBenchmarks` (new) does for the consume path what `MemoryPathBenchmarks` did for the send path: the Memory transport does no serialization, no SQL and no I/O, so what a receive costs there is core library, and a change to it moves every transport.

A receive cost **2,964 ns and 2,833 B**. Three things were paying for nothing.

### 1. Two container resolutions per message

The path resolved `IMessageContext` and `IWorkerNotification` from SimpleInjector for every message. The decisive measurement:

| | time |
|---|---|
| `WorkerNotification`, **resolved** | 437 ns |
| `WorkerNotification`, **constructed** | **20 ns** |

The container is the cost, not the object. Looking the producer up once first does not help either — `InstanceProducer.GetInstance()` measured 666 ns against 690 ns for `Container.GetInstance` — so the answer was to stop resolving, not to resolve faster. Recorded in the harness README so it isn't re-derived.

The factories now build the default implementation directly: `IWorkerNotificationFactory.Create` **437 → 17.6 ns**, `IMessageContextFactory.Create` **823 → 75 ns**.

**A replaced registration still wins.** This matters more than I first thought — my initial read that nothing overrides these was wrong, truncated by a `head` on the grep. SQL Server and PostgreSQL *both* swap in a relational `IWorkerNotification` when the de-queue transaction is held open, and the inbox feature reads that variant back off the context. So the factories ask the container what it would produce and defer to it for anything but the default. The check is a new default interface method `IContainer.GetImplementationType<T>()` returning `null` — "do not know" — so a container implemented elsewhere always defers, and a decorated or conditional registration fails the same safe way. `Defers_To_The_Container_When_The_Registration_Was_Replaced` covers it, and fails if the shortcut is made unconditional.

### 2. Wiring the context allocated twice over

Every transport subscribes commit/rollback/cleanup to each message's context and unsubscribes after. That allocated a delegate per method group at all six sites (384 B), *and*, because `MessageContext` seeded each event with `delegate { }`, a combine on every subscribe and a rebuild on every unsubscribe (313 B). Handlers cached in the transports, seed dropped in the core: **697 B per message → nothing.**

### 3. Cancellation checks

`Tokens.Any(t => t.IsCancellationRequested)` runs twice per message in every transport; `ICancelWork.Tokens` is a `List<T>` and LINQ reaches it as `IEnumerable<T>`, boxing the enumerator each time. New `CancelWorkExtensions.AnyCancellationRequested()` is an indexed loop. The Memory store also rebuilt a linked `CancellationTokenSource` per de-queue from two tokens fixed for the queue's lifetime — 80 B and 58 ns for a result that never changed; built once now.

## Result

| | before | after |
|---|---|---|
| de-queue through the full decorated chain | 2,964 ns / 2,833 B | **2,365 ns / 2,112 B** |
| `IMessageContextFactory.Create` | 823 ns / 277 B | 75 ns / 272 B |
| `IWorkerNotificationFactory.Create` | 437 ns / 85 B | 17.6 ns / 80 B |
| context event wiring | 697 B | 0 B |

Also recorded: **the four receive decorators are cheap** (2,058 B undecorated against 2,112 B decorated, time not separable from noise). The send path's decorators cost ~206 B before #242 fixed them, so that expectation did not carry over — worth knowing before anyone goes looking.

What is left is rebuilding the message itself: of the storage layer's 1,637 B, `IMessageFactory.Create` is 416 B (it copies the header dictionary) and `IReceivedMessageFactory.Create` is 616 B. Neither is a quick win; noted in the README rather than attempted here.

## Tests

- 14 new unit tests: the two factory shortcuts and their fallback, and `AnyCancellationRequested` including the empty and null cases.
- ⚠️ I overwrote `MessageContextFactoryTests.cs` and `WorkerNotificationFactoryTests.cs` instead of adding to them. Caught it before pushing; the original `Create_Default` cases are restored in both.
- Verified non-vacuous: the fallback test fails when the shortcut is made unconditional.
- Green locally: 9 unit projects (2,494), Memory integration (64), Memory LINQ integration (20), LiteDb integration (100). `NoTests.sln -c Release -p:CI=true` clean. SQL Server, PostgreSQL, Redis and SQLite integration need Jenkins — and the SQL Server and PostgreSQL runs are the ones that exercise the replaced-registration path, so they matter here.

Closes #235.

## Summary by CodeRabbit

* **Performance**
  * Improved message consumption efficiency by reducing allocations during context creation, event handling, cancellation checks, and memory-transport operations.
  * End-to-end consumption improved from approximately 2,964 ns / 2,833 B to 2,365 ns / 2,112 B per message.

* **Compatibility**
  * Custom container registrations continue to be honored when creating message contexts and worker notifications.

* **Documentation**
  * Added benchmark coverage and documented findings for memory-transport receive performance.


blehnen:
Sonar and both review comments addressed in 70cdc921.

**The disposal race — a real one, and mine.** Building the linked `CancellationTokenSource` once is what created it: a de-queue now reads its token inside the read lock, while `Dispose` was freeing it from another thread with no synchronisation. The previous per-message source was function-local, so that lifetime didn't exist before this PR. Disposal now takes the **write** lock first — the same guarantee `Clear` already relies on to know no de-queue is in flight — and stays ordered before the lock's own disposal.

**Sonar quality gate** (`new_reliability_rating` 3, one MAJOR `S8949`): `BlockingCollection.Add` in the benchmark's `[IterationSetup]` left the `CancellationToken` overload implicit. Now passes `CancellationToken.None` explicitly, which is the rule's own second option — seeding a fixture is deliberately not cancellable, and the collection is unbounded so the add never blocks. I didn't reshape the floor rung to dodge the rule: it mirrors the `BlockingCollection` + `ConcurrentDictionary` the transport actually uses, and swapping in a plain queue to satisfy a linter would have quietly changed what the floor measures.

**License headers** added to the three test files. Same call as #242: test projects here are mostly without one, but CLAUDE.md asks for it on all source files.

Re-ran after the change: 1,078 unit tests and the full Memory integration suite (64) green, `NoTests.sln -c Release -p:CI=true` clean. The Memory suite is the one that exercises `DataStorage` disposal repeatedly, so it's the relevant check for the fix above.

blehnen:
`codecov/patch` was failing at **87.66% against a 90% floor** — and worth flagging, `gh pr checks` doesn't list Codecov at all, so it wasn't visible alongside the other checks. Only the check-runs API shows it. Fixed in ec3cb10b.

**Not by shaving the diff.** My first instinct was that the cosmetic de-indent in `GetNextMessage` had inflated the patch — 106 of that file's 143 changed lines were pure re-indentation. But the arithmetic says otherwise: those lines were *mostly covered* (39 hits of 52), so removing them would have dropped the numerator faster than the denominator, landing around 84%. I tried it, measured it, and reverted it.

The real gap was 12 uncovered lines and 7 partials. Covered the ones that are genuinely mine, three of which back claims this PR's release notes make and which nothing was verifying:

- **`IContainer.GetImplementationType`'s default body** — reached through a container that doesn't override it. That is the "a container implemented elsewhere always defers" promise; it had no test.
- **`ContainerWrapper.GetImplementationType`** — registered, delegate-registered, and unregistered. The delegate case is the one SQL Server and PostgreSQL depend on.
- **`MessageContext` raising commit/rollback with nothing subscribed** — precisely the branch the removed `delegate { }` seed used to make unreachable, plus the subscribed case so both halves are pinned.
- **`DataStorage`'s poison path** — a body that can't be turned back into a message is reported rather than retried, because the de-queue has already taken it off the queue. Six lines, no prior coverage. Plus disposal with and without a de-queue having built the cancellation source.

Also dropped a null-check that can't fire: an `InstanceProducer` the container hands back always has a `Registration`, so only the producer itself can be missing. That turned a partial branch into a full one.

Measured locally with Coverlet: **145 of 154 changed lines**, ≈94%. What's still bare are two defensive branches in the de-queue — a `KeyNotFoundException` and a missing collection — that can't be reached without corrupting the store's static state. I'm not contriving tests for those.

Full sweep re-run after the change: 9 unit projects (2,506 tests) green, `NoTests.sln -c Release -p:CI=true` clean.

Issue #229, opened 2026-08-27
SQLite integration tests silently leak database files on the 0.6.6.x line


blehnen:
## Summary

On the `release/0.6.6.x` line, the SQLite integration suite leaves real SQLite database files behind in the test output directory — and does so **silently**, without tripping the warning that `IntegrationConnectionInfo` emits when it cannot delete one.

This is pre-existing, not introduced by the 0.6.6.3 backport (#228). It is filed separately because that PR *improves* it substantially but does not eliminate it.

## Measurements

All on net48, same machine, `Source/DotNetWorkQueue.Transport.SQLite.Integration.Tests/bin/Debug/net48`:

| branch | filter | tests | leaked `.db3` | warnings emitted |
|---|---|---|---|---|
| `release/0.6.6.2` | `SimpleProducer` | 144 | **47** | 0 |
| `release/0.6.6.3` | `SimpleProducer` | 144 | **0** | 0 |
| `release/0.6.6.3` | *(full suite)* | 196 | **26** | 0 |

The backported teardown fix from #217 — `SQLiteConnection.ClearAllPools()`, deleting the `-wal`/`-shm`/`-journal` siblings, and a retry — clears the producer tests entirely. Something in the remaining ~52 tests still leaks.

For comparison, the same suite on `master` (net10, 200 tests) leaks **0**.

## What is known

- The leaked files are **real SQLite databases with content**, 40 KB to 1.2 MB — not empty stray files.
- **No warning is printed.** `IntegrationConnectionInfo.Dispose()` logs `WARNING: could not delete test database '...'` when `File.Delete` throws. Zero such lines appear, so every delete it attempted appeared to succeed.
- All 25 `new IntegrationConnectionInfo(...)` sites are inside `using` blocks, so it is not an undisposed fixture.
- It is **not** the in-memory connection string. I tested `FullUri=file:NAME.db3?mode=memory&cache=shared;Version=3;` directly against System.Data.SQLite.Core 1.0.115.5: it creates no file on disk.

## Leading hypothesis

Something recreates the database file *after* the fixture's teardown has deleted it — most likely a consumer, heartbeat or task-scheduler thread that outlives the test body and reopens the connection. That would explain a successful delete, no warning, and a file present afterwards.

This is a hypothesis, not a finding. I ruled out three other explanations before arriving at it and did not get as far as confirming this one.

## Suggested next step

[...]

PR #245, opened 2026-09-03
test(benchmarks): close out the LiteDb pass — concurrency and allocation, both falsified (#234)

### LINKED ISSUE #234: Performance pass: LiteDb transport

blehnen:
Closes #234. **Benchmark and documentation only — no production code changes** (`git diff master --name-only` touches nothing outside `DotNetWorkQueue.Benchmarks`).

Both remaining questions on the LiteDb pass had a hypothesis pointing at a win. Neither survived measurement, and in both cases the thing that was wrong was the *measurement*, not the transport.

## 1. Do concurrent consumers scale? No — and the lock is not why

`ReceiveMessageQueryHandler` holds a process-wide `static` lock. The send path had a lock of exactly that shape and removing it was the largest single finding of this whole pass, so the obvious expectation was that the same win sat here.

| rung | mean | vs 1 thread |
|---|---|---|
| 200 de-queues, 1 thread, one queue | 66.70 ms | 1.00 |
| 200 de-queues, 4 threads, one queue | 80.36 ms | **1.21× slower** |
| 200 de-queues, 8 threads, one queue | 81.92 ms | **1.23× slower** |
| 200 de-queues, 4 threads, two separate queues | 67.59 ms | 1.01× |
| raw correct claim, 4 threads | 21.14 ms | — |
| raw correct claim, 1 thread | 11.56 ms | — |

**The control is the finding.** My first version of it took no lock, and reported the raw engine getting *faster* with four threads (7.4 ms against 21.1 ms) — which would have made the transport's lock look like pure overhead. It was wrong: `BeginTrans` does not block in direct mode, so the claims interleaved. I added a verification and it caught it — **200 claims left only 63 of 200 rows claimed.** The rung ran fast because most of the work was double-delivery.

So a correct claim in LiteDB direct mode *has* to be serialised, which is exactly what the transport's lock does. It is the correctness mechanism, not removable overhead. The remaining degradation is inherent: a correct raw claim also goes 11.6 → 21.1 ms (1.83× worse) with no transport in the way.

Per-database lock keying is the only lever left, and it is the one #238 had to remove — `Path.GetFullPath` does not resolve symlinks, and the failure mode is silent double delivery. Recorded, not attempted.

## 2. Where does the ~50 KB per send go? Nowhere — the floor was wrong

The status comment claimed a direct send allocates ~107 KB against ~47 KB for "the same raw LiteDB writes", leaving ~50 KB unexplained. It was not the same writes:

- the floor wrote the **256-byte payload**; the transport writes the **serialized message**, 4,464 bytes for that same payload
- the floor's meta collection had **no indexes**; the transport's carries **four**

| rung | allocated |
|---|---|
| raw, 256 B body, no indexes (the old floor) | 46.9 KB |
| raw, serialized-size body | 59.8 KB |
| raw, + structural indexes (key, unique `QueueId`) | 72.8 KB |
| raw, + `Status` and `HeartBeat` as well | **114.2 KB** |
| DotNetWorkQueue send, end to end | **101–115 KB across runs** |

The transport sits **at or slightly below its own floor**. There is no transport overhead to find; a send costs what LiteDB charges for the writes it does.

What the ladder does reveal is worth having: **index maintenance is about half a send, and 41.5 KB / 36 µs of it is the `Status` + `HeartBeat` pair.** #241 measured what dropping those costs reads — heartbeat monitor 83 µs → 2.2 ms, 26× worse — but not what keeping them costs writes, which is *every send*. Both sides are now on the record. Neither is adjustable: both options are hard-coded `true`.

## 3. Two smaller items, answered

- **BSON mapper / per-operation connection** — from the code, not a benchmark. `new LiteDatabase(...)` uses `BsonMapper.Global`, which caches its entity mapper per type, and `LiteDbConnectionManager.GetDatabase` holds one `LiteDatabase` for the manager's life in direct mode. Neither is per-message work; only shared mode constructs per operation, which was already measured.
- **Deletion and file-handle behaviour** — no leak on this line. A full 100-test LiteDb run left **zero** new files; the 35 databases in the test output are dated 2021 to 2026-03, all older than this work. But the helper deletes only the main file (not LiteDB's `-log.db`) and swallows a failed delete after one 3-second retry, so a future regression would be silent — the same shape as #229, and worth hardening there rather than here.

## Notes on the harness

- Extracted `ConsumerInternals` so the Memory and LiteDb receive suites share one way of reaching a consumer's container instead of duplicating the reflection.
- The new suite is deliberately **not** a `[MemoryDiagnoser]` one: with one invocation per iteration the diagnoser counts the per-iteration fixtures — two queues, six containers, sixteen receive chains, 400 seeded messages — as de-queue cost, and read 98 MB per 200 de-queues that way.
- Warm-up matters more than usual here. The first benchmark in the process ran 223 ms on its first iteration and 63 ms by its twelfth; five warm-ups were not enough and made the baseline look bimodal rather than warming. Fifteen fixes it.
- Every rung guards against measuring nothing: a de-queue that finds no message throws, and the raw control verifies it claimed every row exactly once.

## Summary by CodeRabbit

* **Benchmarks**
  * Added LiteDB benchmarks for serialized payload writes, indexed writes, structural-index writes, and concurrent dequeue performance.
  * Added comparisons covering one or multiple queues and raw database operations.
  * Added validation that concurrent claims process each row exactly once.
  * Improved benchmark setup for more reliable consumer and resource handling.

* **Documentation**
  * Updated concurrency benchmark results and expanded LiteDB performance findings, including indexing and status/heartbeat trade-offs.


blehnen:
All three addressed in e296bda0.

**Duplicate detection — the right catch for this suite.** The rungs guarded against a de-queue finding *nothing*, but not against two finding the *same thing*, which is the single failure this whole PR is about. A regression in the de-queue lock would still have shown 200 successful claims and passed. Each rung now verifies it claimed exactly `TotalMessages` **distinct** messages, so the rows are a positive statement that the de-queue stays exclusive under four and eight consumers and across two queues — not just a timing.

Worth recording how that went: the first version keyed on the message id alone, and the two-queue rung immediately failed. Not a real double claim — a message id is an auto-increment int scoped to its own database, so queue A and queue B each have an id 1. The key is now queue *and* id. The guard caught its own bug before it caught anything else.

The added dictionary insert shifts every rung uniformly, so the conclusion is unchanged but the numbers move:

| rung | mean | vs 1 thread |
|---|---|---|
| 1 thread, one queue | 68.02 ms | 1.00 |
| 4 threads, one queue | 83.23 ms | 1.23× slower |
| 8 threads, one queue | 85.01 ms | 1.25× slower |
| 4 threads, two separate queues | 70.90 ms | 1.04× |
| raw correct claim, 4 threads | 21.71 ms | — |
| raw correct claim, 1 thread | 11.88 ms | — |

The README is updated to these rather than left at the earlier run's.

**Meta document shape** — added `QueueProcessTime`, `ExpirationTime` and `Route` so the field count matches what the transport writes. I did **not** add indexes for them, and they stay null: the options that populate and index them (delayed processing, message expiration, routing) are all off by default, and the default configuration is what these rungs compare against. `MetaDataTable` only builds those three indexes when the matching option is enabled, so indexing them here would have made the floor charge for work the measured send isn't doing — which is the exact error this PR exists to correct.

**`Fixture.Dispose`** — `Interlocked`-guarded, matching the convention the library uses throughout.

Issue #232, opened 2026-08-28
Performance pass: PostgreSQL transport


blehnen:
The SQLite transport had a full performance pass in 0.10.0 (single send ~8.5 ms → ~80 µs; empty-queue dequeue 29.9 µs / 22.1 KB → 5.8 µs / 648 B). None of the other transports have been measured. This issue covers PostgreSQL.

**Nothing here has been measured yet.** This is an investigation issue: establish where the time and allocation actually go before proposing a change.

## Why PostgreSQL is shaped differently from SQLite

- A network round trip dominates, as with SQL Server, so connection lifecycle is unlikely to be the story. Npgsql pools by default.
- PostgreSQL already has a true bulk-insert batch path from 0.9.41 (`INSERT … SELECT … FROM unnest(@Bodies, @Headers) WITH ORDINALITY`), which sidesteps the bound-parameter limit entirely.

## The PostgreSQL-specific lever

**Npgsql's automatic statement preparation** (`Max Auto Prepare` / `Auto Prepare Min Usages`) has no equivalent in the other providers, and is off by default. Statement preparation was worth roughly 2.4 µs per statement on SQLite, and the send path issues the same handful of statements over and over — exactly the shape auto-prepare targets. Worth measuring first, since if it works it is a connection-string change rather than code.

Note the counterpart risk: prepared statements are per-connection and are invalidated by DDL on the referenced tables, so measure a queue that is created and dropped as well as a steady-state one.

## Other things worth looking at

- **Round trips per send**, and whether any can be collapsed.
- **The decorator stack** — retry, trace and metrics wrap every `ICommandHandlerWithOutput` across all three relational transports. A win here is shared, not transport-local.
- **`Transport.RelationalDatabase` shared code** — check whether any of the SQLite work belongs in the shared layer where all three relational transports would benefit.
- **The dequeue script and its parameters** — built per dequeue, or cached per shape?
- Binary vs text parameter handling for the body `bytea`.

Integration tests need a PostgreSQL instance and a connection string in the integration test project.

## Method

[...]

Issue #231, opened 2026-08-28
Performance pass: SQL Server transport


blehnen:
The SQLite transport had a full performance pass in 0.10.0 (single send ~8.5 ms → ~80 µs; empty-queue dequeue 29.9 µs / 22.1 KB → 5.8 µs / 648 B). None of the other transports have been measured. This issue covers SQL Server.

**Nothing here has been measured yet.** This is an investigation issue: establish where the time and allocation actually go before proposing a change.

## Why SQL Server is shaped differently from SQLite

The SQLite findings do not transfer directly, and assuming they do would waste the pass:

- A network round trip dominates a SQL Server operation. The ~800 ns of core-library overhead removed in #230 was already noise here, and micro-optimisations that mattered at 80 µs will not show at millisecond scale.
- `Microsoft.Data.SqlClient` pools connections by default, so the single largest SQLite win (pooling was off) does not exist here.
- SQL Server already has a true bulk-insert batch path from 0.9.41 (`MERGE … OUTPUT`, ids recovered in caller input order).

So the plausible levers are **round-trip count**, **command and parameter construction**, and the **dequeue script**, not connection lifecycle.

## Specific things worth looking at

- **Round trips per send.** How many does a single `Send` make, and can any be collapsed? This is likely the dominant term.
- **The decorator stack.** The relational transports wrap every `ICommandHandlerWithOutput` in retry, trace and metrics decorators. Measure what the stack itself costs per operation — it is shared by SQL Server, PostgreSQL and SQLite, so a win here is not transport-local.
- **`Transport.RelationalDatabase` shared code** — the same layer SQLite went through. Some of the SQLite work landed in the transport rather than the shared layer; check whether any of it belongs lower down where all three relational transports would get it.
- **The dequeue script and its parameters** — built per dequeue, or cached per shape?
- Whether the async path is genuinely async end to end, or sync work wrapped in a task.

Integration tests need a SQL Server instance and a connection string in the integration test project.

## Method

The SQLite pass (0.10.0) worked by building a benchmark *ladder* rather than profiling: each rung isolates one layer, and the difference between two adjacent rungs is the cost of what the upper one adds. The harness is `Source/DotNetWorkQueue.Benchmarks` (`SendPathBenchmarks`, `ReceivePathBenchmarks`, `CorePathBenchmarks`); its README documents the rungs and, importantly, the hypotheses that measurement killed. Add rungs for this transport rather than starting over.

Ladder to reproduce, per path:

[...]

PR #246, opened 2026-09-03
perf(sqlserver): cache the de-queue statement for routed consumers, and the meta insert SQL (#231)

blehnen:
First tranche of #231. The issue stays open — the biggest measured lever is deliberately not in here (see *What is not in this PR*).

## The floor had to come first, and it changed the plan

The issue predicted round-trip count would dominate. It does not, and saying so required a rung I only wrote after the first numbers looked wrong:

| rung | mean | allocated |
|---|---|---|
| round trip only — `SELECT 1`, pooled | **496 µs** | 5,576 B |
| raw: 1 insert, pooled | 8,491 µs | 7,954 B |
| raw: DNWQ shape, **4 round trips** | 9,429 µs | 16,594 B |
| raw: DNWQ shape, **1 round trip**, same work | 8,284 µs | 9,610 B |
| **transport send, end to end** | **9,654 µs** | **30,522 B** |
| pooled connection open + close | **1.8 µs** | 914 B |

**A write costs seventeen times a round trip here**, and **the library adds 2.3% of a send's time** (225 µs of 9,654). Time optimisation on this path is close to pointless — the database is 97.7% of it. Allocation is the target: 30,522 B against a 16,594 B floor. The pool is free at 1.8 µs, so the single largest SQLite win does not exist here, exactly as the issue said.

## 1. The de-queue statement for routed consumers — the real win

Already cached, so the SQLite finding (script generation = 91% of de-queue allocation) does not transfer. **But the cache was conditional:** `GetDeQueueCommand` returned cached text only with no routes and no user clause. A consumer using either rebuilt the whole statement — a table variable, a CTE, forty-odd appends — on **every poll**, which is the one loop in the library that never stops.

| | before | after |
|---|---|---|
| statement, routed | 438.9 ns / **5,368 B** | 40.9 ns / **80 B** |

**98.5% less allocation per poll**, and 5,368 B was 45% of everything an empty de-queue allocated. The no-routes path is unchanged at 11 ns / 0 B.

Safe to key on because only the route *count* reaches the text — routes become `@Route1..@RouteN` placeholders and their values are bound as parameters — and the user clause is inlined so the clause is part of the key. Both are fixed for a consumer's life, so this is one cache entry per consumer shape, not one per poll.

**The trap was the parameters.** They are not part of the text, so returning cached SQL without also producing them hands a user-clause consumer a statement referencing `@p1` with nothing bound. `A_Cached_Statement_Still_Returns_The_User_Parameters` covers it and was verified to fail — *"the cached path dropped the user parameters"* — when the assignment is removed.

This reverses a deliberate deferral whose note pointed at a `CONCERNS.md` no longer in the repo; @blehnen confirmed it was time, not a hazard. **The identical note is in PostgreSQL's `ReceiveMessage`, so the same change is available under #232.**

## 2. The meta insert SQL — smaller than it first looked

Cached per table-and-option shape. A send goes 30,522 B → 29,298 B: **1,224 B, about 4%.**

I had reported this as 16%. That was wrong and the PR body says so on purpose: the rung measures `BuildMetaCommand` as a whole, and only 1,224 B of its 4,986 B is the text — the rest is `SqlParameter` construction, which no cache removes. The README now warns against re-reading that rung as SQL-generation cost.

The cache key is a maintenance contract: a flag added to `AddBuiltInColumns` but not to `GetMetaSqlShape` would give two option shapes one cached string, i.e. silently wrong SQL. `Every_Option_Shape_Produces_Its_Own_Sql` walks all 32 combinations asserting 32 distinct statements, and was verified to fail naming the colliding shapes when a flag is dropped from the key.

## What is not in this PR

- **Collapsing the send's four round trips into one — 1,145 µs (12%) and ~7 KB.** The largest measured lever. It moves the transaction from client to server and touches the held-transaction inbox fork, the job-exists check and error handling. It needs its own change and a decision on the transaction contract.
- **Parameterising the delay and expiration**, which are written into the SQL as literals (`DATEADD(ms,12345,...)`). That would extend the send-side cache to every message *and* stop SQL Server compiling a fresh plan per distinct delay value.
- The shared `Transport.RelationalDatabase` decorator stack, still unmeasured.

## Two mistakes the ladder made first, both recorded

Both produced plausible-looking numbers, which is why they are in the README:

- It reported round trips as not worth collapsing **from a floor it had not measured** — with no `SELECT 1` rung there was nothing to say whether 9 ms was a round trip or a write.
- It reported the held-connection rung as the **slowest** (14.1 ms vs 8.4 ms pooled for identical work). The rungs insert and never delete, BenchmarkDotNet runs them in declaration order, and later rungs paid for the data-file growth earlier ones caused. Truncating per iteration removed it.

## Tests

- 10 new unit tests across the two caches, both non-vacuity-checked as described above.
- Green: 180 SQL Server unit tests, 78 producer + 42 route/consumer integration tests against a real instance, `NoTests.sln -c Release -p:CI=true` clean.
- One production-visible extra: `InternalsVisibleTo("DotNetWorkQueue.Benchmarks")` on the SQL Server transport, so the harness measures `BuildMetaCommand` directly instead of through reflection in a hot loop — a mistake this harness has already had to correct once. Say the word if you would rather not widen that.

## Summary by CodeRabbit

- **Performance**
  - Improved SQL Server transport performance by caching reusable statements during message sending and receiving.
  - Reduced memory allocations for routed message polling and standard message sends.
  - Preserved dynamic handling for delayed and expiring messages.

- **Documentation**
  - Added SQL Server performance benchmarks and measurement guidance covering sends, receives, batching, connections, and statement generation.
  - Documented routed-consumer caching as a regression guard.


blehnen:
Good catch, and it's a labelling bug the fix itself introduced. Fixed in 2dc63292.

Before routes were cached, `Statement_Rebuilt` did exactly what its name said — routes bypassed the cache, so it reassembled the statement on every call. That is the measurement that justified this PR. **After the fix, the same rung measures a cache hit for the routed shape**, so the name and description were describing behaviour that no longer exists.

Renamed to `Statement_Routed` / *"statement: routed consumer, no round trip"*, and the doc comment now records both readings, because the rung's meaning changed underneath the number:

- rebuilding (before): 439 ns, 5,368 B
- cache hit (now): 41 ns, 80 B — the 80 B is the composite key

Re-measured after the rename to confirm it still reports the hit: 36.5 ns / 80 B against 10.4 ns / 0 B for the unrouted path.

I kept the rung rather than deleting it, and said why in the comment: it is the **regression guard**. If the route cache ever stops working, that row goes straight back to kilobytes per poll and says so. A rung that forces a genuine rebuild is not reachable through this API without inventing a new route count on every call — which would grow the cache without bound, the exact thing the key is designed to prevent.

The README's finding row now carries the same note, so the 5,368 B → 80 B figure can't be misread as two different rungs.

blehnen:
Both addressed in 4aa980b5 — and both are the same stale wording as the rename, which I fixed on the method but not on the class or the setup.

**Stale remarks.** The class-level `<remarks>` and the setup comment still said routes bypass the cache and that these rungs measure a rebuild. That stopped being true with the fix in this branch. They now say what the rungs actually measure — cache hits for two different shapes — and why they are kept rather than deleted: they are the regression guard for the keying.

**Unwarmed routed key.** A fair catch with a real consequence. `Setup` primed only the unrouted key, and the routed key is a separate entry, so the routed rung's *first* call did the build it claims not to measure. BenchmarkDotNet's warm-up almost certainly absorbed it — but "almost certainly" is not what a rung's description should rest on, especially in a suite whose whole point this PR is. Both keys are primed now, so each rung measures a hit by construction.

Re-measured after the change, unchanged as expected:

| rung | mean | allocated |
|---|---|---|
| statement: cached (no routes) | 10.3 ns | 0 B |
| statement: routed consumer | 37.5 ns | 80 B |

Issue #159, opened 2026-06-23
Replace FluentAssertions with MSTest assertions (license: 6.12.2 is last MIT release)


blehnen:
## Background

`FluentAssertions` is pinned at **6.12.2** in `Source/Directory.Packages.props` — the last MIT-licensed release. Version 7.x and later moved to a **paid/commercial (non-OSS) license**, so we cannot upgrade without taking on a license obligation.

Because of the pin, every dependency-refresh release (0.9.37, 0.9.39, …) has to deliberately hold FluentAssertions back and document the exception. `dotnet list package --outdated` flags `6.12.2 → 8.x` on essentially every test project, which adds noise to each refresh.

## Proposal

Replace FluentAssertions usage across the test projects with the built-in **MSTest assertions** (`Assert.*`, `CollectionAssert.*`, `Assert.ThrowsException<T>`, etc.), then drop the `FluentAssertions` package entirely. The solution already standardized on MSTest 3.x/4.x as the test framework, so this removes a third-party assertion dependency rather than adding one.

### Scope
- All test projects currently referencing `FluentAssertions` (Tests + IntegrationTests + Dashboard test projects).
- Remove the `FluentAssertions` `<PackageVersion>` from `Directory.Packages.props` once no references remain.
- No production-code impact.

### Considerations
- High file count but mechanical; `.Should().Be(x)` → `Assert.AreEqual(x, actual)`, `.Should().BeTrue()` → `Assert.IsTrue(...)`, `.Should().Throw<T>()` → `Assert.ThrowsException<T>(...)`, collection/contains variants → `CollectionAssert` / LINQ + `Assert`.
- Some FA fluent chains (approximate equality, complex object-graph comparison) have no 1:1 MSTest equivalent — those spots may need `CompareNETObjects` (already a dependency) or small custom helpers.
- Can be done incrementally per test project.

## Acceptance
- No `FluentAssertions` references remain in the solution.
- `FluentAssertions` removed from `Directory.Packages.props`.
- All test suites pass (unit + integration) on Jenkins.

PR #247, opened 2026-09-03
perf(postgresql): cache the de-queue statement for routed consumers (#232)

blehnen:
First piece of #232. The issue stays open — its own headline lever (Npgsql auto-prepare) is not measured yet.

## The same defect, a different number

PostgreSQL's `ReceiveMessage.GetDeQueueCommand` carried the *identical* deferral note to the one fixed for SQL Server in #231, pointing at the same `CONCERNS.md` that no longer exists in the repo. Same shape: the cached statement was returned only when there were no routes and no user clause, so a routed consumer reassembled it on **every poll** — the one loop in the library that never stops.

I measured it here rather than carrying SQL Server's figure across, and that was the right call:

| rung | before | after |
|---|---|---|
| statement: cached (no routes) | 10.1 ns / **0 B** | unchanged |
| statement: routed consumer | 240.5 ns / **2,648 B** | **37.1 ns / 80 B** |

**2,648 B against SQL Server's 5,368 B — about half.** PostgreSQL uses an updating CTE with `FOR UPDATE SKIP LOCKED` rather than a table variable, so its statement is smaller. Real either way, and a **97% cut** once cached. Transferring the other transport's number would have overstated this by 2×.

## Why the key is safe, and bounded

Only the route *count* reaches the statement text — routes become `@Route1..@RouteN` placeholders and their values are bound as parameters — and the user clause is inlined, so the clause itself is part of the key. Both are fixed for a consumer's lifetime, so this is one cache entry per consumer shape, not one per poll.

**The parameters were the trap here too.** They are not part of the text, so returning cached SQL without also producing them hands a user-clause consumer a statement referencing `@p1` with nothing bound. `A_Cached_Statement_Still_Returns_The_User_Parameters` covers it and was verified to fail — *"the cached path dropped the user parameters"* — when the assignment is removed.

## Not in this PR

- **Npgsql auto-prepare**, which #232 names as *the* PostgreSQL-specific lever and which is off by default. If it works it is a connection-string change rather than code. Unmeasured so far, and it needs the counterpart risk checked too: prepared statements are per-connection and are invalidated by DDL on the referenced tables, so a queue that is created and dropped has to be measured as well as a steady-state one.
- Round trips per send, the shared decorator stack, and binary vs text `bytea` handling.

## Tests

- 5 new unit tests mirroring the SQL Server set: per-route-count statements, cache hits for the same count, no-routes not confused with routes, the `userParams` trap, and two user clauses not sharing a statement.
- Green: 166 PostgreSQL unit tests, 67 route/consumer integration tests against a real instance, `NoTests.sln -c Release -p:CI=true` clean.
- The routed rung changed meaning with the fix — it rebuilt before, it is a cache hit now — so its remarks say both, and it is kept as the regression guard. That is the correction CodeRabbit caught on #246; this suite carries it from the start.
- `InternalsVisibleTo("DotNetWorkQueue.Benchmarks")` added to the PostgreSQL transport, as on SQL Server, so the harness measures `GetDeQueueCommand` directly rather than through reflection in a hot loop.

## Summary by CodeRabbit

- **Performance**
  - Improved PostgreSQL and SQL Server dequeue efficiency by reusing statements for each route count.
  - PostgreSQL benchmark results show a reduction in cached statement key size from 2,648 bytes to 80 bytes—approximately 97%.

- **Bug Fixes**
  - User-defined clauses are now evaluated on every dequeue, ensuring changing filters and parameters are honored correctly.
  - User-defined clauses are excluded from caching to prevent unbounded cache growth.


blehnen:
Thanks @qodo-code-review — finding 3 is correct and serious, and I've fixed it in ecf6dbba. Findings 1 and 2 I'm declining, with reasons below.

## 3. Dynamic clauses grow cache — **valid, High, and mine**

Confirmed by reading the API rather than taking it on trust: `SetUserParametersAndClause` takes a `Func<string>`, and `GetUserClause` **invokes it on every de-queue**. So a caller's clause is free to differ each time, and keying the cache on its text adds a permanent entry per poll to a cache that lives as long as the consumer. That is unbounded growth ending in memory exhaustion.

It was in **both** transports — this PR and **#246, already merged** — so both are fixed here. My PR descriptions for both called the cache "bounded"; that was wrong and the README and changelog now say so.

The fix: the clause is no longer part of the key, and a statement carrying one is not cached at all — which is exactly the behaviour those consumers had before this work. **The route win is untouched**, because the routed measurements never involved a user clause: only the route count is keyed on, and counts are small integers.

Worth flagging how the regression test had to be written, because my first attempt was wrong. The obvious test — *a changing clause returns the current text* — **passes on the buggy code**, since a per-clause key produces a different key each call and so still builds fresh. The bug is growth, not staleness. The test that actually catches it asserts on the cache itself, that the keys a per-clause implementation would have written do not exist. Verified: it fails on the restored buggy key while the behavioural test passes.

This is the same hazard the send-side cache in #246 was explicitly written to avoid — delay and expiration are not cached *because* they are per-message literals — and I walked into it anyway through a different door.

## 1. MSTest 3.x — declining

The repo is on MSTest **4.2.3** via central package management (`Directory.Packages.props`), and has been since the FluentAssertions removal work. These tests match every other test in the project; there is no per-project override available under CPM, and downgrading one project is not a change I can make in isolation. If the checklist still says 3.x, the checklist is stale rather than the code.

## 2. FluentAssertions — declining

FluentAssertions was **deliberately removed from this repository** in #159, across eight PRs, on licensing grounds — 6.12.2 was the last MIT-licensed release. Reintroducing it here would reverse a completed migration. MSTest assertions are the standard in this codebase now.

Both are worth checking against the repo's own history before applying; I'd rather say so than quietly comply.

Re-verified after the fix: 182 SQL Server and 168 PostgreSQL unit tests green, `NoTests.sln -c Release -p:CI=true` clean. Integration re-run is next.

PR #248, opened 2026-09-03
docs(postgresql): auto-prepare measured and documented, deliberately not enabled (#232)

blehnen:
Second piece of #232, and the one the issue named as its headline lever. **No behaviour change** — the outcome is a measurement, a permanent test, and documentation.

## Auto-prepare pays, modestly

`Max Auto Prepare` is off by default in Npgsql, has no equivalent in the other providers, and the send path issues the same handful of statements repeatedly — the shape it targets. If it paid, it would be a connection-string change rather than code, which is the cheapest kind of win available.

| operation | off (as shipped) | on | gain |
|---|---|---|---|
| single send | 5.938 ms ± 0.317 | 5.565 ms ± 0.071 | ~6% |
| batch of 100 | 37.386 ms ± 0.399 | 32.460 ms ± 0.410 | ~13% |

Allocation is unchanged. **The batch figure is the reliable one**; the single-send gain is about one error bar wide on the `off` rung and is written up as suggestive rather than firm.

## A number I nearly reported

A dry run of this suite showed **55.8 ms against 10.5 ms — a five-fold win.** It was nonsense: one invocation per rung, with the `off` fixture running first and paying the connection and warm-up cost for both. The same order-dependent trap the SQL Server ladder had to correct earlier in this sweep. It is in the benchmark README because it looked spectacular and meant nothing.

## Why it is documented rather than enabled

Prepared statements live on the physical connection, the pool hands that connection back out, and dropping or recreating a table invalidates any prepared statement referencing it — and this library creates and drops queues as a normal operation. Turning auto-prepare on by default would hand that exposure to every PostgreSQL consumer in exchange for 6–13%. That is not a trade to make on a user's behalf, so the transport README describes it as a tuning option with both numbers and the caveat.

`AutoPrepareSurvivesDdl` drives the risky path permanently: create → send past the prepare threshold → drop → recreate **under the same queue name**, so the SQL text is identical and any prepared statement is stale → send again, single and batch. It passes.

**What that test does not do is prove DDL-safety.** Npgsql exposes no public counter for auto-prepared statements, so I cannot verify preparation actually happened before the drop. It is evidence the scenario works, not proof the invalidation path was exercised — and the test, the commit and the README all say so rather than letting a green tick imply more.

## Still open on #232

Round trips per send, the shared `Transport.RelationalDatabase` decorator stack (a win there would also count toward #231), and binary vs text `bytea` handling.

## Summary by CodeRabbit

* **Documentation**
  * Added PostgreSQL performance-tuning guidance for Npgsql automatic statement preparation.
  * Documented how to enable the option, its trade-offs, and measured results: no measurable difference for single sends and 13–17% faster batches of 100 messages.
  * Clarified that automatic preparation remains disabled by default because queue recreation can invalidate prepared statements.
  * Added benchmark findings and coverage confirming sends continue working after a queue is recreated.


blehnen:
Addressed in 75b262fe.

**Baseline may enable auto-prepare** (@qodo-code-review, #4) — valid, and it changed the result. `Setup` passed the ambient connection string to the `off` rung unchanged, so a caller whose string already enabled auto-prepare would have had a baseline that was not off. Both rungs now state their setting explicitly: `Max Auto Prepare=0` against `Max Auto Prepare=20`.

Re-measured on that footing, **the single-send gain is not real**:

| operation | effect |
|---|---|
| single send | no measurable difference (7.174 ms vs 7.365 ms, ±0.3) |
| batch of 100 | 13–17% faster across two runs |

The transport README, benchmarks README and changelog are corrected to say the benefit is on the batch path and only there. The previous wording would have sent someone to change a connection string for a gain that isn't there.

**License header** (both reviewers) — added.

**Fence language** (@coderabbitai) — set to `text`.

**MSTest 3.x and FluentAssertions** — declining, as on #247. The repository is on MSTest 4.2.3 under central package management and a single project cannot opt out of it; FluentAssertions was deliberately removed in #159 on licensing grounds, so adding it back here would reverse a completed migration.

Re-verified: 168 PostgreSQL unit tests, `AutoPrepareSurvivesDdl` still passing, `NoTests.sln -c Release -p:CI=true` clean.

Issue #264, opened 2026-09-07
Integration test time: measured — it is a few clock-bound consumer tests, not the DataRow matrices


blehnen:
The integration stages dominate a Jenkins build (~30-40 minutes of a ~45 minute run). This asks whether that time is buying anything, or whether some of it is the same ground covered repeatedly.

**It may well be fine.** The suite is broad on purpose and this project has been bitten before by a transport-specific defect that only a per-transport test would have caught (Redis history). Nothing here should be cut because a table looks repetitive — the point of the exercise is to find out whether the repetition is load-bearing, and the honest outcome may be that it is.

## What the suite currently is

**1,473 expanded test cases** across the integration stages (a `DataRow` counts as one case):

| project | cases |
|---|---|
| Dashboard.Api | 408 |
| SQLite | 200 |
| PostgreSQL | 181 |
| SqlServer | 150 |
| LiteDB | 100 |
| SqlServer.Linq | 83 |
| PostgreSQL.Linq / SQLite.Linq | 80 / 60 |
| Memory | 64 |
| Redis / Redis.Linq | 48 / 28 |
| LiteDB.Linq / Memory.Linq | 47 / 20 |
| TaskScheduler.Distributed | 4 |

From the message-count column alone that is on the order of **147,000 messages** sent across a full run — 100 rows push 1,000 messages each, 280 push 100, 293 push 10.

## Two obvious explanations, already ruled out

Recording these so nobody spends the afternoon rediscovering them.

**The `.Linq` projects are not duplicates of the plain ones.** Comparing which shared scenario classes each drives, the overlap per transport is **1-2 classes**, and those are helpers (`MsTestHelper`, `FakeMessage`) rather than scenarios. The `.Linq` projects drive the method and compiled-expression producer/consumer surface, which the plain projects never touch.

**Running the same shared scenario once per transport is the design, not an accident.** `SimpleProducer`, `SimpleConsumer`, `ConsumerErrorTable` and friends are each driven by 6 transport projects, which is exactly what `CLAUDE.md` asks for: *"Features that might vary by the transport implementation should have integration Tests; This has caused issues before with Redis History for example."*

## Where the multiplication actually comes from

The producer matrices, which are a product of independent flags. SQLite's `SimpleProducer`:

```
9 option combinations  ×  inMemoryDb (2)  ×  enableChaos (2)  =  36 rows
```

[...]

Issue #260, opened 2026-09-05
Jenkins runner sizing: stages compile 42 projects when they need ~6, and containers have no memory limit


blehnen:
## Symptom

A Docker runner sits at **99.94% CPU** shortly after a build starts — measured on a 16-core / 24-thread i9-12900K with 128 GiB RAM, of which only 23 GiB was in use and temperatures were well inside limits. Nothing is starved except CPU, and the spike lines up with the start of the parallel integration stage rather than with the tests themselves.

**Builds are green, so this is not urgent.** Filing it so the cause is written down rather than rediscovered, and because the obvious reaction — reduce the number of runners — is the wrong one: that lengthens wall-clock time directly, and the machine has capacity to spare on every axis except CPU.

## Cause: every parallel stage compiles the whole solution

`Jenkinsfile` contains **15** occurrences of:

```sh
dotnet build "Source/DotNetWorkQueue.sln" -c Debug
```

One in stage 1, then one in each of the ~14 parallel integration stages. That solution is **42 projects**, and nothing shares build output between containers — the only `stash`/`unstash` calls carry coverage XML and JUnit results, so each container compiles all 42 from scratch, at the same time.

Measured against the actual project-reference closures, each stage needs far less than it builds:

| stage | projects it needs |
|---|---|
| Dashboard.Api.Integration.Tests | 12 |
| SqlServer.Linq / PostgreSQL.Linq / SQLite.Linq | 7 |
| SqlServer / PostgreSQL / Redis.Linq / SQLite / LiteDB.Linq | 6 |
| Redis / LiteDB / Memory / TaskScheduler.Distributed | 5 |
| Memory.Linq | 3 |

**Mean 6.1 projects of 42 — each stage compiles about 6.8x what it needs, or roughly 502 unnecessary project-builds per run.**

Among those 42 are `Dashboard.Ui` (Blazor, 20 `.razor` files), `Dashboard.Ui.Tests` and `Dashboard.Ui.E2E.Tests`. Razor is the most CPU-hungry compilation in the tree and no transport integration stage references it — it is compiled 14 times per run for nobody.

There is also no MSBuild parallelism cap anywhere in the Jenkinsfile, so each of the 14 containers defaults to one build node per core and tries to take the whole host at once.

## Two levers, cheapest first

[...]

PR #269, opened 2026-09-07
Build each stage's own test project instead of the whole solution

blehnen:
Addresses the build half of #260.

All fourteen parallel stages ran `dotnet build Source/DotNetWorkQueue.sln`, compiling **42 projects when the mean stage needs about 6**. Every stage compiled `Dashboard.Ui` — Blazor, the most CPU-hungry compilation in the tree — for nobody.

## Measured, not estimated

Clean solution build against a clean single-project build, same machine, same state:

| | time |
|---|---|
| full solution | 19.64 s |
| one integration project | **3.49 s** |
| | **5.6x, 82% less** |

Jenkins measures its solution build at 1m40s per stage, so this is roughly **1m22s saved per stage** — about **19 minutes of agent time** across fourteen stages, and **~1.4 minutes off the critical path** (set by the three stages at 32–34 min).

**That wall-clock number is deliberately unexciting, and worth being straight about.** The issue's framing implied the compile was the dominant cost; the #264 measurement showed it is about 5% of a stage. The reason to do this anyway is that #264 also found the real lever — a few clock-bound consumer tests, `ConsumerHeartbeat` alone being 25% of a critical-path stage — and shortening their timers safely needs margin against a loaded agent. This buys that margin.

## What to look at

**The risk in this change is runtime, not compile.** Building only the test project changes what lands in the output directory, so anything loaded by reflection rather than referenced would compile fine and fail at *run* time — a worse failure than the one being fixed.

`Dashboard.Api` does exactly that: `Assembly.LoadFrom` in `DashboardExtensions`, and a two-stage `Type.GetType` → `Assembly.LoadFrom` resolver in `DashboardService`. It probes `AppContext.BaseDirectory` and configured plugin paths, **not** sibling project outputs — and `Dashboard.Api.Integration.Tests` references all six transports explicitly, so they land in its own bin.

Verified rather than reasoned: all fifteen projects build standalone, and the Dashboard API integration tests pass **219/219 on both net10.0 and net8.0** from a standalone build. The one test that configures a plugin path uses an empty temp directory, so it does not depend on a sibling build either.

**Stage 1 is untouched** — it runs eleven test projects and genuinely needs the solution.

## Not done here

The MSBuild parallelism cap (`-m:`) the issue also proposes. Landing both at once would make the effect of either unattributable, and with 42 projects down to 6 the cap may no longer be needed — better to measure this first.

The memory half of #260 — container cap sizing on the 32 GiB host and a `--memory` limit in the agent template — is Jenkins configuration rather than repository content, so it stays with the issue.

## Summary by CodeRabbit

* **Chores**
  * Updated integration-test pipeline stages to build only their respective test projects.
  * Improved build targeting across database, in-memory, dashboard, and distributed task-scheduler integration tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,7 +102,7 @@ pipeline {
                     steps {
                         sleep(time: 0, unit: 'SECONDS')
                         catchError(buildResult: 'FAILURE', stageResult: 'FAILURE') {
-                            sh 'dotnet build "Source/DotNetWorkQueue.sln" -c Debug'
+                            sh 'dotnet build "Source/DotNetWorkQueue.Transport.SqlServer.IntegrationTests/DotNetWorkQueue.Transport.SqlServer.Integration.Tests.csproj" -c Debug'
                             withCredentials([string(credentialsId: 'sqlserver-connstring', variable: 'SQLSERVER_CONN')]) {
                                 sh 'echo "$SQLSERVER_CONN" > "Source/DotNetWorkQueue.Transport.SqlServer.IntegrationTests/bin/Debug/net10.0/connectionstring.txt"'
                             }
@@ -124,7 +124,7 @@ pipeline {
                     steps {
                         sleep(time: 5, unit: 'SECONDS')
                         catchError(buildResult: 'FAILURE', stageResult: 'FAILURE') {
-                            sh 'dotnet build "Source/DotNetWorkQueue.sln" -c Debug'
+                            sh 'dotnet build "Source/DotNetWorkQueue.Transport.SqlServer.Linq.Integration.Tests/DotNetWorkQueue.Transport.SqlServer.Linq.Integration.Tests.csproj" -c Debug'
                             withCredentials([string(credentialsId: 'sqlserver-connstring', variable: 'SQLSERVER_CONN')]) {
                                 sh 'echo "$SQLSERVER_CONN" > "Source/DotNetWorkQueue.Transport.SqlServer.Linq.Integration.Tests/bin/Debug/net10.0/connectionstring.txt"'
                             }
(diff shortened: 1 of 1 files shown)


blehnen:
Jenkins is green on this PR, and the result is better than the estimate above — but the estimate was right for the wrong reason, so the numbers are worth restating.

**Observed:** Integration Tests **34m → 31m**. The three long stages moved 32–34m → PostgreSQL 31m, SQLite Linq 31m, SQLite 30m. `Build & Unit Tests` is untouched at 7m13s, as expected.

I predicted ~1.4 min. The likely reason it is closer to 3 is that the isolated 19.64s solution build is not what Jenkins was paying: fourteen agents each compiling 42 projects at once is precisely the CPU saturation #260 was filed about, so the per-stage build cost under contention was higher than the uncontended measurement. That is an explanation, not a proof — it is one build against one baseline, and stage times carry real noise.

**More useful: the shape of the run has changed.**

| | |
|---|---|
| total agent time | 267.9 min across 15 stages |
| wall clock | 31 min |
| average concurrency | 8.64 |
| longest stage | 31 min (PostgreSQL) |
| spread | 1m56s … 31m |

Wall clock (31) now equals the longest stage (31) exactly. That rules out being executor-bound: at capacity 8 the floor would be 267.9/8 = **33.5 min**, above what was observed, so capacity is at least 9 and scheduling is near-perfect. **The build is bound by its single slowest stage, not by how much total work there is.**

That matters for #264, and it sharpens its target. Shaving total agent time is now worth almost nothing in wall clock; only the three 30–31m stages count. `ConsumerHeartbeat` at 3 cases × 142s is 25% of the SQLite stage — directly on the critical path — which is the case for doing the timer work.

It also puts a ceiling on stage-splitting as an alternative: halving all three long stages would only move the wall to SqlServer Linq at 28m.

Issue #256, opened 2026-09-05
No async receive path in core: IReceiveMessages is sync-only


blehnen:
Raised by #231, which asked whether the SQL Server async path is "genuinely async end to end, or sync work wrapped in a task". It was never examined during the performance pass. Recording it because the answer is already known to be structural, and it is not a SQL Server question.

## The shape of it

There is no asynchronous receive path in the core. `IReceiveMessages` is synchronous only, and `MessageProcessingAsync` calls the synchronous `ReceiveMessage` — so an "async" consumer is synchronous receive work wrapped in a task, on every transport, not just the relational ones.

That means:

- the async consumer occupies a thread pool thread for the duration of the database round trip rather than releasing it
- under load the pool is the limiting factor rather than the database, which is the same shape as the Redis sync-concurrency finding in #161
- no transport can fix this on its own — it needs `IReceiveMessagesAsync` in the core first, with each transport then implementing it

## Why it matters beyond performance

Any future "deprecate the synchronous API" work depends on this. The sync API cannot be deprecated while it is the only real receive path — the async surface is a wrapper over it.

## Not scheduled

Larger than a performance pass and touches every transport. Filing so that closing #231 does not lose it.

blehnen:
**Release-note requirement — recording here because there is nowhere else it will survive.**

PR #279 (PR 1 of 7) is a **breaking change**: `IWaitForEventOrCancel` and `IWaitForEventOrCancelThreadPool` gain abstract members. Both are public, and `ATaskScheduler.WaitForFreeThread` exposes the latter, so any custom `ATaskScheduler` implementation breaks on rebuild.

No changelog entry was added in that PR. `CHANGELOG.md` has no `Unreleased` section, and the 0.11.0 precedent is that entries are written in one pass during the release PR — creating a section would have been a unilateral convention change. So the requirement lives here instead, on the issue that stays open until PR 7 and that the release notes will be written from.

**When this ships, the changelog needs a ⚠️ bullet along these lines:**

> ⚠️ Custom `ATaskScheduler` implementations must implement the new async members on `IWaitForEventOrCancel` and `IWaitForEventOrCancelThreadPool`. The built-in schedulers are unaffected.

Expect the later PRs to add more: the six transport interfaces each gain an async member, which breaks any transport implemented outside this repository.

---

**Also worth its own issue, found while doing PR 1** — not blocking anything:

[...]

Issue #161, opened 2026-06-23
Migrate Redis transport to StackExchange.Redis 3.x (thread-pool starvation on sync Lua path)


blehnen:
## Background

Bumping `StackExchange.Redis` 2.13.17 → **3.0.0** in the 0.9.39 dependency refresh (PR #160) caused intermittent **operation timeouts** in the Redis integration suite (e.g. `RouteMultiTests.Run`, `SimpleProducer.Run`). It was reverted to 2.13.17 for that release; this issue tracks doing the 3.x migration properly.

## Symptom

```
Timeout performing SCRIPT/EVAL (15000ms), ... rs: ReadAsync, ws: Idle, in: 0, last-in: 33,
sync-ops: 104-122, ... clientName: ...(SE.Redis-v3.0.0.11234),
WORKER: (Busy=79..99, Min=64, Max=32767),
POOL: (Threads=99, QueuedItems=4-6, CompletedItems=...)
```

at `EnqueueLua.Execute` → `RedisDatabase.ScriptEvaluate` → `ExecuteSync`.

## Root cause (analysis)

This is **.NET thread-pool starvation**, not a slow/dead connection:
- The connection is healthy and the reply **already arrived** (`last-in: 33ms`, `in: 0`, `rs: ReadAsync`).
- `WORKER: Busy=79..99` against `Min=64` with `POOL.QueuedItems` backing up — the global thread pool is saturated.
- The transport executes Lua **synchronously** (`BaseLua.TryExecute` → `ScriptEvaluate` → `ExecuteSync`). Each concurrent sync EVAL parks a worker thread waiting on the reply; with no free thread to complete the reply, the 15s sync timeout fires.

**Why it's new in 3.0:** StackExchange.Redis 3.0 reworked socket I/O scheduling toward the .NET global thread pool. Per the SE.Redis docs, `SocketManager.Shared` uses a dedicated ~10-thread pool while `SocketManager.ThreadPool` uses the global pool *"with caution due to potential thread pool starvation."* The 2.x default insulated this synchronous-heavy workload; 3.0 exposes it.

Related: 3.0 also defaults to **RESP3** (vs RESP2 in 2.x) — not the cause here, but worth validating during migration.

## Options to evaluate

[...]

PR #270, opened 2026-09-07
Run integration tests in parallel within each assembly

blehnen:
Addresses #264. Integration is **32 of a ~40 minute build**; this takes it to a projected **~12 minutes**.

## The cause was not what the issue assumed

#264 proposed trimming the producer `DataRow` matrices. Measurement killed that: they are **15%** of SQLite's stage time and **6%** of PostgreSQL's. The heartbeat and monitor intervals were already short.

The real shape is in the shared consumer scenarios — elapsed time is `Thread.Sleep(runtime) × ceil(messageCount / workerCount)`. The stages were **asleep**, and both Jenkins hosts confirm it: **1.71% and 0.75% CPU** during the test phase, 6.4 GiB of 32 and 10.7 GiB of 128 in use.

MSTest does not parallelize within an assembly unless told to. Nothing in the repo told it to, so every stage ran its cases one at a time while the agent idled.

## Measured

Locally, off the drvfs mount, against the same PostgreSQL the agents use:

| | cases | sequential | 4 workers | | peak RSS |
|---|---|---|---|---|---|
| SQLite | 200 | 27m15s | **7m39s** | **3.56x** | 1.93 GiB |
| PostgreSQL | 181 | 29m53s | **8m13s** | **3.64x** | 2.18 GiB |

200/200 and 181/181 pass.

**The number to look at is not the speedup.** It is the sum of per-test durations, which rose **1.2%** and **0.5%**. The tests did not start contending — they stopped queueing. Scenario by scenario the clock-bound ones are unchanged (`ConsumerHeartbeat` 0.99x, `ConsumerCancelWork` 0.98x); only the short producer tests, which do real work, show any cost, and none of it large.

## Projection against the current master build

| stage | now | projected | | stage | now | projected |
|---|---|---|---|---|---|---|
| SQLite Linq | 32 | **12.0** | | LiteDB / LiteDB Linq | 18 | 6.7 |
| PostgreSQL | 31 | 11.6 | | Dashboard | 16 | 6.0 |
| SQLite | 29 | 10.9 | | Redis Linq / Memory Linq | 12-13 | ~4.7 |
| SqlServer Linq | 28 | 10.5 | | Redis / Memory | ~10 | ~3.7 |
| SqlServer / PostgreSQL Linq | 24 | 9.0 | | | | |

**Integration 32 → ~12 min; whole build ~40 → ~20 min.**

Applied to all thirteen stages rather than the slow three deliberately: parallelising only the top would drop them to ~11 and leave SqlServer Linq at 28 as the new ceiling — 32 to 28 for all that work.

## Why four workers

A connection limit, not CPU or memory. One stage holds **~59** connections to PostgreSQL, and three stages share that server in CI (PostgreSQL, PostgreSQL Linq, Dashboard.Api) against `max_connections=500`. Eight workers would approach 470.

## What to watch on this build

Local runs cannot reproduce CI: Coverlet is on there and was not here, containers bound the memory, and three stages hit each server concurrently. That is what this run is for, and stages fail independently so any problem names itself.

- **Dashboard.Api** is the least-proven — 408 cases, and the only suite standing up web hosts.
- **Redis and LiteDb** were not run locally at all.
- `MsTestHelper.ClearSynchronizationContext()` runs once on one thread at `[AssemblyInitialize]`, so parallel threads are not covered. Harmless at four workers on both suites tested; it is the first thing to suspect if something hangs.

Excluded: **TaskScheduler.Distributed** already declares `DoNotParallelize`, and the Playwright **E2E** stage at ~2 minutes is not worth the risk.

Not included, deliberately: the `runtime=45` fix for SQLite's consumer scenarios (a 2016 value that missed the 2017 scale-back — five other transports run 15). It is a clean follow-up, and bundling it would confound the duration measurement this PR exists to produce. Both are written up in `docs/lessons-learned.md`.

## Summary by CodeRabbit

* **Tests**
  * Integration test suites now run test methods in parallel, reducing overall execution time.
  * Redis integration tests use a two-worker limit for improved stability.
  * Metric verification checks allow more time for slower test environments.

* **Documentation**
  * Added performance results and guidance for integration test parallelization.
  * Documented historical differences in SQLite runtime settings.


blehnen:
Four builds: two green, two failed. Pushed `ce06cbec` addressing both failures, which turned out to be unrelated to each other — **and neither is the shared services host.**

## The host is not the problem

Stage times across all four runs:

| stage | #1 | #2 | #3 | #4 | spread |
|---|---|---|---|---|---|
| PostgreSQL | 9.3 | 9.7 | 9.6 | 9.3 | 4% |
| Redis Linq | 6.7 | 6.9 | 6.6 | 6.9 | 4% |
| SqlServer | 7.3 | 8.0 | 7.3 | 7.9 | 10% |

**#1 and #2 are the failed builds, and they were not the slow ones.** A shared server under strain produces broad drifting slowdown across everything that touches it, not two isolated tests in otherwise-normal runs.

## Redis (#1): thread-pool starvation, client side

`Timeout performing SCRIPT (15000ms)` — and the request never left the client: `qu: 0, qs: 0, aw: False, in: 0, in-pipe: 0`, against `WORKER (Busy=215, Free=32552, Min=200, Max=32767)`.

Past `MinThreads` the pool adds threads at one or two a second. Every worker is blocked in a synchronous Redis call waiting on a completion that itself needs a worker thread. They deadlock until injection catches up, which is past `syncTimeout`.

That is **#161 — "Migrate Redis transport to StackExchange.Redis 3.x (thread-pool starvation on sync Lua path)"**. `SCRIPT` is that path.

The timings had already said so. Parallel efficiency across the thirteen stages:

```
PostgreSQL      3.33x      LiteDB Linq   2.61x
SqlServer Linq  3.32x      Memory        2.55x
SQLite Linq     3.20x      Dashboard     2.26x
SQLite          3.19x      Redis Linq    1.88x   <- worst of thirteen
```

Redis was already at its concurrency ceiling before this PR pushed it over. **Two workers** halves the concurrent sync calls; Redis Linq should land near 9 min against the 10 min SQLite Linq sets as the critical path, so it costs no wall clock. The real fix is an async receive path (#256).

## PostgreSQL (#2): a poll timeout, not a lost message

`MultiConsumerAsync (25,1,180,10,5,5,True,True)` — a chaos row — reported `Expected 25, Actual 24` from `VerifyMetrics.PollUntil`. Nothing was lost; the last counter had not caught up within the 15s poll window. That helper exists specifically to absorb the gap between a handler signalling completion and the metric recording it, and its own comment says the 15s default was sized for "CI load" — measured when suites ran one test at a time.

Raised to 30s across all six entry points. A real failure still fails, 15s later.

## Not fixed here: the retry never ran

Every stage passes `-- --retry-failed-tests 1`. That is a **Microsoft.Testing.Platform** flag; these projects run under **VSTest** (no `EnableMSTestRunner` — blocked since PR #89). VSTest parses everything after `--` as RunSettings `Key=Value` pairs and silently discards it. No error, no warning.

Both of these flakes would most likely have been absorbed by the retry we believed was running. Worth its own issue — the proper fix is the MSTest runner migration.

## Where the numbers stand

| | before | after |
|---|---|---|
| Integration | 32.0 min | **10.0 min** |
| whole build | ~40 min | **~18.3 min** |

**2.18x, 22 minutes off every build.**

Issue #271, opened 2026-09-08
Enable the MSTest runner so test retries actually work


blehnen:
Six Jenkins stages pass `-- --retry-failed-tests 1`. It does nothing. Every transient failure in those stages has been a red build, and a full rebuild, for months.

## Proof

A one-test project, MSTest 4.2.3 and Microsoft.NET.Test.Sdk 18.7.0 exactly as pinned here, whose only test always fails and appends a line to a file each time its body runs:

```
run WITHOUT the flag
  Failed! - Failed: 1, Passed: 0, Total: 1
  test body executed: 1 time(s)

run WITH -- --retry-failed-tests 1   (byte for byte what the Jenkinsfile does)
  Failed! - Failed: 1, Passed: 0, Total: 1
  test body executed: 1 time(s)
```

Once, either way. **No error, no warning** — which is why this went unnoticed.

## Why

`--retry-failed-tests` belongs to **Microsoft.Testing.Platform**. These projects run under **VSTest**: no `EnableMSTestRunner` or `TestingPlatformDotnetTestSupport` anywhere in the tree, and every run banners `VSTest version 18.0.2 (x64)`.

`dotnet test` passes everything after `--` to the runner. VSTest expects RunSettings as `Key=Value` pairs, so `--retry-failed-tests` and `1` match nothing and are discarded silently.

## How we got here

- **PR #89**, *"ci: enable MSTest runner for test retry support"* — closed, reverted. `EnableMSTestRunner` is all-or-nothing across the repo and **28 projects** reference `MSTest.TestAdapter`.
- **`eb84367c`** (2026-03-31), *"...add test retry for remote transports"* — added the flag to six stages. Without the runner from #89, it was inert from the first commit.

So the intent was right and the prerequisite never landed.

## Why it matters now

#264's parallelism work made this expensive. Four builds of #270 produced two failures, and both were the kind a single retry absorbs:

- **Redis** — StackExchange.Redis sync Lua path thread-pool starvation (#161). Load-dependent by nature.
- **PostgreSQL** — a metric counter 1 short of 25 inside a 15s poll window, on a chaos row.

Neither indicates a defect in the library. Both cost a full build. With one build running at a time, a spurious red is among the more expensive things that can happen here.

## Options

[...]

PR #274, opened 2026-09-08
Stagger longest stage first, and drop the retry flag that never ran

blehnen:
Two small Jenkinsfile changes. Neither alters what runs.

## 1. Stagger the longest stage first — 35s, free

Slots were assigned in source order, so the critical-path stage sat in a late one:

| | was | now |
|---|---|---|
| SQLite Linq (9m25s of work — the critical path) | 35s | **0s** |
| SqlServer (7m20s) | 0s | 25s |
| Dashboard UI E2E (59s of work) | 70s | 75s |

Only the slot occupied by the *longest* stage affects wall clock. Sorting by measured work:

| | now | after |
|---|---|---|
| parallel block | 600s (10m00s) | **565s (9m25s)** |
| whole build | 668s | **633s (10m33s)** |

**The spacing and spread are unchanged** — still 5s apart, still 0–75s. The request pacing to nuget.org is byte-for-byte identical, so nothing about the rate-limit protection the stagger exists for is weakened.

Ordering is by measured work from the last green master build; if stage times shift materially it is worth re-sorting, but only the top three or four slots matter.

## 2. Remove `-- --retry-failed-tests 1` — it has never run

Six stages passed it. It does nothing.

`--retry-failed-tests` belongs to **Microsoft.Testing.Platform**. These projects run under **VSTest** — no `EnableMSTestRunner` anywhere in the tree, and every run banners `VSTest version 18.0.2`. VSTest reads what follows `--` as RunSettings `Key=Value` pairs, so it is discarded with no error and no warning.

Demonstrated rather than assumed, with a one-test project pinned to the same MSTest 4.2.3 and Test.Sdk 18.7.0, whose only test always fails and appends a line each time its body runs:

```
without the flag:  Failed: 1  ->  test body executed 1 time
with the flag:     Failed: 1  ->  test body executed 1 time
```

**How it got here:** PR #89, *"ci: enable MSTest runner for test retry support"*, was reverted — `EnableMSTestRunner` is all-or-nothing across what is now 28 projects. The flag was added afterwards in `eb84367c` and was inert from its first commit. Right intent, missing prerequisite, no error to reveal it.

Removing it does not remove protection. It removes the *appearance* of protection, which is worse than none — two of the four #270 validation builds failed on transient faults a working retry would have absorbed. #271 tracks getting real retries; the fix there is the MSTest runner migration, which is its own project.

## Verification

No Groovy or Jenkins linter is available locally, so syntax is validated by the run this triggers. Statically: brace balance 163/163 unchanged, all 16 stages re-staggered exactly once, 26 `dotnet test` invocations intact, zero `sh` blocks left with a dangling line-continuation backslash, no BOM, LF preserved.

## Summary by CodeRabbit

- **Tests**
  - Updated CI timing for unit, integration, dashboard, task-scheduler, and end-to-end test stages.
  - Unit and integration tests now run without automatically retrying failed tests.
- **Chores**
  - Preserved existing build, reporting, artifact-stashing, and credential-handling behavior.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -47,8 +47,24 @@ pipeline {
 
         stage('Tests') {
             parallel {
-                // Last in the clone stagger: this branch is ~7 minutes against a
-                // 10-minute critical path, so a late start costs nothing.
+                // About the sleeps below. They pace nuget.org, not git.
+                //
+                // A stage-level agent does an implicit checkout before its steps run, and
+                // nothing here sets skipDefaultCheckout, so every sleep lands *after* that
+                // stage has already cloned and *before* its `dotnet build`, which is where
+                // the restore happens. Sixteen restores arriving together is what
+                // nuget.org rate-limits, and that is what the spacing prevents.
+                //
+                // So the clones do all go out at once, and always have. That has never
+                // been the thing that fails, which is why this is left as it is. If clone
+                // pacing is ever wanted, it needs options { skipDefaultCheckout() } plus
+                // an explicit checkout scm after the sleep, in all sixteen stages.
+                //
+                // Slots run longest-work-first so the critical-path stage starts at zero.
+                // Only the slot the longest stage occupies affects wall clock.
+                //
+                // This branch is ~3 minutes of work against a 9m25s critical path, so its
+                // late slot costs nothing.
                 stage('Unit Tests') {
                     agent { label 'docker' }
                     steps {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,7 +52,7 @@ pipeline {
                 stage('Unit Tests') {
                     agent { label 'docker' }
                     steps {
-                        sleep(time: 75, unit: 'SECONDS')
+                        sleep(time: 65, unit: 'SECONDS')
                         catchError(buildResult: 'FAILURE', stageResult: 'FAILURE') {
                             sh 'dotnet restore "Source/DotNetWorkQueue.sln"'
                             sh 'dotnet build "Source/DotNetWorkQueue.sln" -c Debug --no-restore'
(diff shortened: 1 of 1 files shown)


blehnen:
Both findings are valid. One is fixed; the other is real but deliberately not acted on, and the reason is now recorded in the file.

## CodeRabbit — the sleeps do not stagger clones. Correct.

Verified rather than taken on trust: there is no `skipDefaultCheckout` and no explicit `checkout scm` anywhere in the file, so every stage-level agent does an implicit checkout before its steps run. The Jenkins stage view shows the order outright:

```
✓ Check out from version control      4s
✓ Sleep
✓ dotnet build "Source/...csproj"
✓ echo "$SQLSERVER_CONN" > ...
✓ dotnet test  ...
```

Checkout first, sleep second. So the clones do all go out together, and always have.

**What the sleeps actually pace is the restore inside `dotnet build`** — and that is the thing this stagger was added for. nuget.org rate-limits sixteen simultaneous restores and the restores fail; that is the failure this prevents, and it prevents it correctly. The mechanism works. Only my description of it was wrong, and the PR body inherited that error from a stale note.

**Not restructuring, for three reasons:**

1. The clones have been simultaneous for the entire life of this pipeline and have never been what fails. There is no observed problem to fix.
2. Moving the sleeps ahead of the checkout means `options { skipDefaultCheckout() }` plus an explicit `checkout scm` in all sixteen stages — a large change to buy protection against something that has never happened.
3. The stagger is about to be re-evaluated wholesale. blehnen/dotnetworkqueue-ci#2 bakes a read-only NuGet fallback folder into the CI image, which removes the nuget.org pressure entirely. Once that is published the spacing may not be needed at all, and restructuring it now would be work thrown away.

Both the finding and the recipe are written into the comment so the next person does not re-derive it.

## Qodo — stale comment. Correct, and mine.

The Unit Tests comment still said "last in the clone stagger" while this PR moved it to 65s with two branches behind it at 70 and 75. Fixed in `99ee9b53`, along with the "clone" language which was wrong for the reason above.

Thanks both — the checkout-ordering point in particular was worth knowing, and it corrects a note that had been wrong in this repo's records for a while.

PR #275, opened 2026-09-08
Restore from an internal NuGet mirror when the controller offers one

blehnen:
Sixteen agents restoring the same 146 packages at once is what nuget.org rate-limits, and a rate-limited restore **fails the stage**. That is the reason the parallel stages are staggered five seconds apart. A local mirror removes the cause rather than pacing around it.

## The hostname is not committed, and that is the point

`NUGET_SOURCE` is a global environment variable on the Jenkins controller. It resolves only on the internal network and this repository is public, so it stays out of the tree.

**When it is unset, nothing is written and restores go to nuget.org exactly as before** — an external contributor, a fork, or a local reproduction is unaffected. That is why this is a conditional helper rather than a `NuGet.Config` committed to the repo.

The file lands at the **workspace root** rather than a user profile because NuGet searches ancestor directories of the project being restored. That makes it independent of whichever uid the Docker Cloud plugin gives the container — a `~/.nuget` config would not be.

## Verified before committing

The mirror is BaGet in upstream-mirror mode, behind an HTTPS reverse proxy.

| check | result |
|---|---|
| V3 service index | valid, 12 resources |
| upstream mirroring | ✅ 84 / 90 / 112 versions for newtonsoft.json, polly, sqlclient |
| TLS | Let's Encrypt — publicly trusted, **no CA install needed in the container** |
| **full solution restore, cold global + http caches** | **146 packages, 1.8 GB, 23s, 0 errors** |
| push endpoint without an API key | **401** — nothing can be injected into the build supply chain |
| DNS | private address, not internet-exposed |

And both branches of the helper itself:

```
NUGET_SOURCE unset -> no file written, rc 0, restores go to nuget.org
NUGET_SOURCE set   -> config written; `dotnet nuget config paths` finds it
                      from a nested project dir; restore succeeds
```

Worth recording: over plain HTTP this fails outright with `NU1302 — NuGet requires HTTPS sources`. The HTTPS proxy is what makes this work without an `allowInsecureConnections="true"` opt-out in the config.

**Groovy parse checked** in a container, since a Jenkinsfile is otherwise only validated by the run it triggers — and #274 showed how that goes.

## What to look at

The helper runs as the first step of all sixteen parallel stages and the compile gate — seventeen call sites, one line each. `<clear />` means the mirror is the *only* source when it is configured; that is deliberate, since listing nuget.org alongside it would have NuGet querying both and defeat the purpose.

Expect to see this at the top of every stage log:

```
Restores will use the internal mirror at https://…/v3/index.json
```

If that line is absent, the variable is not reaching the agents and restores have silently fallen back to nuget.org — which is a safe failure, just not the intended one.

## Follow-up, deliberately not bundled

With restores coming off a LAN mirror, the 0–75s stagger has little left to do and should be re-measured and probably shrunk. Doing it here would confound the two changes; #260's lesson is that a Jenkinsfile change is validated by the run it triggers, so one change per run.

This also supersedes blehnen/dotnetworkqueue-ci#2, which baked a read-only fallback folder into the CI image. That was the zero-infrastructure answer to the same problem; the mirror is better on every axis — no staleness, 1.6 GB smaller image, and it serves local development too. I would close that PR once this one is green.

## Summary by CodeRabbit

* **Build and Test Infrastructure**
  * Builds and automated test runs now use the configured internal NuGet package source when available.
  * Package restore behavior remains unchanged when no internal source is configured.


blehnen:
Two findings. One was a real bug and is fixed in `bbdbfd29`; the other I am declining, with reasons.

## The stale config — valid, and the premise checks out

Both bots flagged the same thing: the unset branch returned early without removing a `NuGet.Config` an earlier build had written. Since that file carries `<clear />`, the mirror is the *only* source in it — so once `NUGET_SOURCE` went away, every later restore in that workspace would keep pointing at a source that no longer exists, and this PR's central claim (unset ⇒ falls back to nuget.org) would be false.

**The premise it depends on is workspace reuse, and that holds here.** From the Build stage timings:

```
checkout          4s    a fresh full clone of this repo cannot be 4s
dotnet restore    6s    a cold restore of 146 packages (1.8 GB) cannot be 6s
```

Against 23s for a genuinely cold restore measured locally. State plainly persists between builds, so this was reachable.

**Fixed with a marker rather than an unconditional `rm`.** The generated file carries a `do not commit` marker in an XML comment, and the unset branch removes the file only when that marker is present — so a `NuGet.Config` this repository legitimately gains later is left alone. That was the option CodeRabbit suggested as the alternative, and it is the safer of the two.

Verified by extracting the helper's shell body **verbatim** from the Jenkinsfile and executing all four paths against the real mirror:

```
1. set                        -> writes config, marker present
2. unset after set            -> "Removed a stale generated NuGet.Config"   <- the bug
3. unset, hand-written file   -> preserved, not clobbered
4. set again                  -> overwrites; `dotnet nuget config paths` finds it from
                                 a nested project dir; restore succeeds
```

## The missing automated test — declining

Qodo is right that this is conditional logic with no test, running in seventeen stages. I am not adding one, for three reasons:

1. **There is no Jenkinsfile test harness in this repository.** Standing up Jenkins Pipeline Unit to cover ~20 lines of shell would be a larger, riskier change than the thing it tests.
2. **Both paths are exercised, just not by a committed test.** The verification above ran the real shell body, not a paraphrase — including the failure mode this review found.
3. **The failure mode is now safe and self-announcing.** Every build prints which mode it is in (`Restores will use the internal mirror at …`, or `Removed a stale generated NuGet.Config…`), so a misconfiguration is visible in the log rather than silent. And with the fix, the risky path actively self-corrects.

Worth stating plainly rather than hiding: this is still CI code validated by the run it triggers, which is exactly how #274 turned up a problem no static check would have caught. The Groovy parse is checked in a container, but that only rules out syntax.

Issue #162, opened 2026-06-24
True bulk insert for batch Send (Send(List<T>) currently loops single-row inserts)


blehnen:
## Summary

The `IProducerQueue<T>.Send(List<TMessage>)` batch overloads exist but provide **no performance benefit over calling `Send(message)` in a loop** — they do not perform a true bulk/multi-row insert. We should add a real batched insert path, with the biggest payoff for the SQLite (and other relational) transports.

## Current behavior

The batch overloads fan the list out to the **single-message** command handler, one call per message:

- `IProducerQueue<T>` exposes `Send(List<TMessage>)` / `Send(List<QueueMessage<...>>)` (`Source/DotNetWorkQueue/IProducerQueue.cs:45-74`).
- `ProducerQueue.InternalSend(List<...>)` hands the whole list to `ISendMessages.Send(List<...>)` (`Source/DotNetWorkQueue/Queue/ProducerQueue.cs:313`).
- `SendMessages.Send(List<...>)` (`Source/DotNetWorkQueue.Transport.Shared/Basic/SendMessages.cs:79`) runs a **`Parallel.ForEach`** (sync) / plain **`foreach`** (async) that calls the single-message `SendMessageCommand` handler once per message.
- The relational `SendMessageCommandHandler` (e.g. SQLite `Source/DotNetWorkQueue.Transport.SQLite/Basic/CommandHandler/SendMessageCommandHandler.cs:111`) opens **its own connection + transaction per message**, then does 2–3 single-row `ExecuteNonQuery` round trips: insert body (`INSERT INTO Queue (Body, Headers) VALUES (@Body, @Headers)`), `SELECT last_insert_rowid();`, insert metadata row, and optionally a status row.

Net result: `Send(List<N>)` == N independent transactions and N×(2–3) single-row round trips. The sync path opens N concurrent connections. There is no shared transaction and no multi-row insert.

## Why it matters (esp. SQLite)

SQLite is single-writer (see #4). Every transaction is a commit/fsync. Collapsing N message inserts into one transaction with multi-row inserts removes the per-message commit overhead and connection churn, which should be a large throughput win for bulk producers. Relational transports generally benefit (fewer round trips, one transaction), with SQLite benefiting most.

## Proposed approach

Add a true batch send path behind the existing API surface — no change to `IProducerQueue` / `IQueueOutputMessages`:

[...]

PR #277, opened 2026-09-08
Design: async receive path (#256)

blehnen:
Design only — no code. Records what was agreed for #256 before implementation, per the architectural path in the brainstorming skill.

## Two things the issue understated

**The synchronous receive is deliberate.** `MessageProcessingAsync.DoTryAsync` carries a comment saying the call *must* block or the task scheduler's limit is ignored — a single dequeue thread parked in `WaitForFreeThread.Wait(workGroup)` is what caps in-flight work at N. The interface is the easy part; **backpressure is the problem**, and the issue doesn't mention it.

**Receive is not the only blocking call.** `Commit` blocks once per message. The idle back-off blocks one pool thread *per worker* via `WaitHandle.WaitOne`. And the heartbeat blocks too — `SmartThreadPoolTaskScheduler` dispatches with `Task.Factory.StartNew`, so despite the name it owns no threads; `ThreadsMax` bounds in-flight heartbeat tasks, not which pool runs them. That matches #161's evidence exactly: `WORKER: (Busy=215, Min=200)` is the .NET pool.

## Decisions recorded

| | |
|---|---|
| goal | genuine async I/O, not an async surface over sync work |
| scope | six transport interfaces + `IQueueWait` |
| backpressure | awaitable throttle — work groups are a required feature and are preserved |
| API | extend existing interfaces, matching `ISendMessages` which already carries `Send` and `SendAsync` together |
| return type | `ValueTask<T>` + `CA2012` raised to error |

**On `CA2012`:** verified it does **not** fire by default in this repo — a build with both a double-await and a `.Result` block passed. With the rule set to error, both become build errors. Tests would not have caught it: a `ValueTask` over a completed result (every embedded-transport receive, every empty poll) tolerates misuse silently.

## The acceptance test already exists

`StarvationBaselineTests` caps the worker pool to 6, floods 50 senders, and documents its own expected outcome as **RED** — a permanent diagnostic, excluded from CI. An async twin under the identical cap **must go green while the sync one stays red**. Differential, same harness, one variable, and it cannot pass by accident.

## Delivery

Seven PRs split by interface — extending an interface breaks all six transports at once, which rules out a core-first split. **PR 2 (`IReceiveMessages`) is the one that matters and the one to stop at** if the starvation twin won't go green: one PR finds that out rather than seven. One release, gated, per #162's precedent.

Breaking change — added interface members break any external transport — so the changelog needs a ⚠️ bullet.

## What to review

Whether the decisions are the ones you'd want defended in six months. The delivery order is the part I'd most welcome disagreement on: PR 2 is large (one interface, four decorators, six transports, the consumer loop) and there may be a better cut.

## Summary by CodeRabbit

* **Documentation**
  * Added an approved design specification for end-to-end asynchronous receive operations.
  * Documented async transport behavior, cancellation and exception handling, throttling, backpressure, testing requirements, staged delivery, and breaking-change considerations.
  * Clarified that SQLite and LiteDb implementations remain internally synchronous.

### DIFF
--- a/docs/superpowers/specs/2026-09-08-async-receive-path-design.md
+++ b/docs/superpowers/specs/2026-09-08-async-receive-path-design.md
@@ -1,0 +1,192 @@
+# Async receive path — design
+
+Issue: #256. Date: 2026-09-08. Status: approved, not started.
+
+## The problem
+
+`IReceiveMessages` is synchronous only, and `MessageProcessingAsync` calls it. An
+"async" consumer is therefore synchronous receive work wrapped in a task, on every
+transport.
+
+The cost is not slowness, it is a deadlock shape. #161 recorded
+`WORKER: (Busy=215, Free=32552, Min=200, Max=32767)` — every .NET pool thread blocked
+in a synchronous Redis call, waiting on completions that themselves needed a pool
+thread to run. Past `MinThreads` the pool adds threads at one or two a second, so the
+two only unwind when injection catches up, which is past `syncTimeout`. Redis is
+currently capped at `Workers = 2` in the integration suite as a workaround for exactly
+this.
+
+## What the issue understated
+
+**The synchronous call is deliberate.** `MessageProcessingAsync.DoTryAsync` carries:
+
+```
+//this call must block; otherwise, the limitations enforced by the task scheduler will be ignored
+//calling the async method here may result in hundreds of de-queues at once, instead of the N set in the configuration
+```
+
+A single dequeue thread parked in `WaitForFreeThread.Wait(workGroup)` is what caps
+in-flight work at N. Making receive non-blocking removes that cap unless something
+replaces it. The interface is the easy part; **the backpressure model is the problem.**
+
+**Receive is not the only blocking call.** Three of these are on the per-message happy
+path (throttle, receive, commit); the rest are conditional but land on the same pool:
+
+| where | what blocks |
+|---|---|
+| `SchedulerMessageHandler` | `WaitForFreeThread.Wait(group)` |
+| `MessageProcessingAsync.DoTryAsync` | `ReceiveMessage(context)` |
+| `QueueWait.WaitInternal` (idle) | `StopWorkToken.WaitHandle.WaitOne(t)` |
+| `ProcessMessageAsync` | `Commit(context)` |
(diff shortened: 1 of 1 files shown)


blehnen:
Both findings are valid and both are now in the spec (`b930ad92`). I verified each against the code rather than taking them on trust, and both hold.

## The `WaitAsync` cancellation contract

Correct, and the consequence is worse than a style difference.

`WaitInternal` uses `StopWorkToken.WaitHandle.WaitOne(t)`, which **returns normally** when the token fires and lets `Wait` go on to advance the back-off index. `await Task.Delay(t, token)` **throws** instead. The caller is:

```csharp
_noMessageToProcessBackOffHelper.Value.Wait();
return;
```

so a throw would land in the outer `catch (OperationCanceledException)` — which performs a rollback and fires a rollback notification, for a poll that found no message and has nothing to roll back.

The spec now pins the contract explicitly: bind to `ICancelWork.StopWorkToken` and **not** `CancelWorkToken` (the existing code notes it deliberately uses only the stop token, and that the ordering between them is enforced by implementation rather than contract); return rather than throw; advance `_currentIndex` in both cases; return immediately for a zero or already-signalled wait; `QueueWaitNoOp` stays a no-op. With the swallowing `try`/`catch` written out, since that is the non-obvious part.

## Per-transport invariants

Also correct, and this is the more valuable of the two. Verified: `SQLServerMessageQueueReceive` attaches `context.Commit +=` and `context.Rollback +=` — transactional and non-transactional variants — stores the connection on the context and registers a cached cleanup handler. `RedisQueueReceiveMessages` has a `while (true)` loop over expired messages with `workSub.Reset()` on the work signal.

**Receive wires up the commit lifecycle; it is not a pure fetch.** An async rewrite that returns a message and nothing else would compile, return messages, and pass the four `ConsumerAsync*` scenarios while having dropped a cleanup handler or an expired-message retry.

The spec now carries a per-transport invariant table and requires explicit tests for it — commit and rollback handlers firing, headers and message id surviving a poison message, Redis retrying an expired message rather than returning it, the work signal reset. I have flagged it in the spec as the part of this work most likely to break quietly, which is exactly the reason the review was worth having.

Thanks — the second one in particular is a gap I would not have caught until it bit during PR 2.

PR #278, opened 2026-09-08
Plan: PR 1 of the async receive path — the awaitable throttle

blehnen:
Implementation plan for **PR 1 only** of #256. No code.

## Why PR 1 only

The spec calls PR 2 the go/no-go gate — *"the one that matters and the one to stop at if the starvation twin will not go green"*. Planning PRs 3–7 in detail now would be planning past a decision point, and would bake in assumptions about an async-wait API that does not exist yet. PR 2 gets its own plan once this lands.

Each plan still produces working, testable software on its own, which is the bar.

## Three tasks

| task | deliverable |
|---|---|
| 1 | `CA2012` raised to error — **before** any `ValueTask` exists, so the twenty decorators to come are written under the analyzer rather than audited after |
| 2 | `ValueTask<bool> IWaitForEventOrCancel.WaitAsync()` |
| 3 | `ValueTask<bool> IWaitForEventOrCancelThreadPool.WaitAsync(IWorkGroup)` |

Nothing calls the new methods in this PR. It changes no behaviour.

## The part worth reading closely

`ManualResetEventSlim` has no `WaitAsync`, so a `TaskCompletionSource<bool>` sits alongside it. The subtle rule is in `Reset`:

```csharp
var current = Volatile.Read(ref _asyncWait);
if (current.Task.IsCompleted)
    Interlocked.CompareExchange(ref _asyncWait, CreateWait(), current);
```

**Only re-arm a source that has already completed.** Replacing one that still has waiters would orphan them — the next `Set` completes the replacement while they go on awaiting the old one. In this codebase that is a worker which stops dequeuing until something unrelated frees a thread: an intermittent hang, and the worst kind to find in an integration suite.

The tests cover the interleavings specifically rather than the happy path — Set during a pending wait, Reset during a pending wait, Set-then-Reset before the waiter resumes, cancellation mid-wait, fifty waiters released by one Set — and Step 6 runs the class **twenty times**, because an ordering bug will not fail every run.

## Verified while writing, so the implementer does not have to

- **`Assert.ThrowsExactlyAsync<T>` exists in MSTest 4.2.3** — compiled a probe against the real package. `strings` cannot answer this reliably on .NET metadata (it suffix-shares, which has produced a wrong retracted claim in this repo before).
- **`InternalsVisibleTo("DotNetWorkQueue.Tests")` is already declared**, so the two `internal` classes are reachable from the test assembly. Noted in the plan so nobody stops on seeing `internal`.
- **`CA2012` is off by default here** and does fire when set to error — both directions confirmed by build.

## Self-review

Placeholder scan clean; names consistent across tasks (`_asyncWait`, `CreateWait()`, `CreateCompletedWait()`, both `WaitAsync` overloads).

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous waiting APIs for event and cancellation signals.
  * Added work-group-aware asynchronous waiting for thread-pool processing.
  * Supports cancellation, resets, multiple concurrent waiters, and disposal scenarios while retaining existing synchronous behavior.

* **Documentation**
  * Added an implementation plan documenting async receive-path behavior, interface updates, synchronization, edge cases, and verification coverage.

### DIFF
--- a/docs/superpowers/plans/2026-09-08-async-receive-path-pr1.md
+++ b/docs/superpowers/plans/2026-09-08-async-receive-path-pr1.md
@@ -1,0 +1,617 @@
+# Async receive path — PR 1: the awaitable throttle
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Give `IWaitForEventOrCancel` and `IWaitForEventOrCancelThreadPool` an awaitable wait, so a later PR can stop blocking a thread-pool thread to enforce consumer backpressure.
+
+**Architecture:** `WaitForEventOrCancel` wraps a `ManualResetEventSlim`, which has no `WaitAsync`. A `TaskCompletionSource<bool>` is added alongside it, swapped on `Reset` only when already completed so waiters are never orphaned. The sync path is untouched. Nothing calls the new methods in this PR — it is purely additive.
+
+**Tech Stack:** .NET 10 / .NET 8, MSTest 4.2.3, NSubstitute, AutoFixture.
+
+**Spec:** `docs/superpowers/specs/2026-09-08-async-receive-path-design.md`
+
+## Scope
+
+This plan is **PR 1 of the seven** in the spec. PR 2 (`IReceiveMessages`) is the project's
+go/no-go gate and gets its own plan once this lands, when the async-wait API is fixed
+rather than assumed.
+
+## Global Constraints
+
+- Target frameworks are `net10.0` and `net8.0`. Anything written must compile on both.
+- `ValueTask<T>` is the return type for new async methods (spec decision). `CA2012` is
+  raised to error in Task 1 and must stay green.
+- Never use `Task.Run` to make sync work look async.
+- `ConfigureAwait(false)` on every await.
+- LGPL-2.1 header on every new non-test `.cs` file — copy the header from the file being
+  modified. Test files in this repo do not carry it.
+- `.cs` files are CRLF with no BOM (`.gitattributes`: `*.cs text eol=crlf`). Match the
+  file you are editing; verify with `head -c3 <file> | od -An -tx1`.
+- MSTest 4.x has no `Assert.ThrowsException`. Use `Assert.Throws<T>` or
+  `Assert.ThrowsExactly<T>`.
+- Release builds set `TreatWarningsAsErrors`. Check with
+  `dotnet build "Source/DotNetWorkQueueNoTests.sln" -c Release -p:CI=true` before the
+  final commit.
+
+## File Structure
+
+| file | responsibility |
+|---|---|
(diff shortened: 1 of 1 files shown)


blehnen:
Five findings, four fixed in `a9f5874a` and one recorded as an accepted decision. Three were real races, and they share a root cause worth naming: **I used lock-free primitives for a three-part state machine — reset event, completion source, cancellation token — that has to stay coherent.**

## Set/Reset interleaving — the important one

Traced through:

```
T1 Reset:  _resetEvent.Reset()
T1 Reset:  current = A                 (A completed)
T2 Set:    _resetEvent.Set()           <-- event now SIGNALLED
T2 Set:    A.TrySetResult(true)        <-- no-op, A already completed
T1 Reset:  CAS _asyncWait = B          <-- B INCOMPLETE
```

The event ends signalled while the async source does not, so a synchronous waiter proceeds and an async waiter hangs. In PR 2 the consumer loop is async-only, which makes that a worker that stops dequeuing — **exactly the hang the CAS guard was written to prevent, arriving by a route it doesn't cover.**

## Dispose

Correct. An async waiter holds only a `Task`; disposing the primitives underneath doesn't fault it, so a pending `WaitAsync` waited forever on a dead object. The synchronous `Wait` at least throws.

## Cancel then Reset

Correct, and subtle. `Cancel` completes the source; `Reset` sees "completed" and re-arms it, leaving a later `WaitAsync` waiting on an object that can never be signalled again. `Wait` returns `false` forever because it consults the token — so `WaitAsync` now checks it first.

## The fix

Every transition happens under one lock. Completing a source inside it is safe precisely because `RunContinuationsAsynchronously` prevents an inline continuation running while the lock is held — that flag was already there for a different reason and turns out to carry this one too.

Four regression tests added, including a 2000-iteration concurrent `Set`/`Reset` asserting the two views always agree:

```csharp
if (syncSignaled && !asyncTask.IsCompleted)
    Assert.Fail($"attempt {attempt}: sync signaled but async still pending");
```

## Null-group routing

Fair — asserting the initial signalled state would also pass for an implementation giving `null` its own private source. The test now resets, asserts pending, sets, and asserts released, plus a second test that a group `Set` does **not** release the shared waiter.

## Breaking API change — recorded, not changed

Correct that both interfaces are public and `ATaskScheduler` exposes one of them, so a custom scheduler breaks on rebuild.

Not acting on it: the spec already chose to extend existing interfaces rather than add parallel async ones, matching `ISendMessages`, which carries `Send` and `SendAsync` together, and the maintainer accepted the breaking change explicitly. But that decision was taken about the six *transport* interfaces and these two were not named at the time, so the plan now says so and flags that the release note must cover them. If it turns out not to be wanted, the alternative is a separate `IWaitForEventOrCancelAsync` — a second interface and a "which one does this implement" question at every call site.

---

Worth noting for anyone reading the thread: one of the tool outputs in the review asserted that `IWaitForEventOrCancel` does not exist in this library and that cancellation is handled via `CancellationToken` on `IWorkerNotifications`. That appears to come from public documentation rather than this repository — the type is at `Source/DotNetWorkQueue/IWaitForEventOrCancel.cs`. The findings themselves were all checked against the actual code and four of them held.

PR #279, opened 2026-09-08
Awaitable throttle for the consumer backpressure primitive (#256, PR 1 of 7)

blehnen:
First of the seven PRs in the [async receive path design](docs/superpowers/specs/2026-09-08-async-receive-path-design.md), implementing [its plan](docs/superpowers/plans/2026-09-08-async-receive-path-pr1.md).

**Nothing calls the new methods.** This PR changes no behaviour — it adds the awaitable wait a later PR needs so the consumer loop can stop blocking a thread-pool thread on every dequeue. Blocking there is a deadlock shape, not merely slow: a blocked pool thread cannot run the continuation that would release it, which is what #161 recorded as `WORKER: (Busy=215, Min=200)`.

## What's here

| commit | |
|---|---|
| `faacd550` | `CA2012` raised to error, **before** any `ValueTask` exists |
| `40227405` | `ValueTask<bool> IWaitForEventOrCancel.WaitAsync()` |
| `4c286f7f` | `ValueTask<bool> IWaitForEventOrCancelThreadPool.WaitAsync(IWorkGroup)` |
| `eb7a36af` | fix: lost wake-up + stop allocating on the scheduler path |
| `af030c7a` | fix: `Dispose`/`WaitAsync` race that could hang a first-ever waiter |
| `ebe6a529` | test: replace a vacuous 50-second concurrency test |

## ⚠️ Breaking change

`IWaitForEventOrCancel` and `IWaitForEventOrCancelThreadPool` are public, and `ATaskScheduler.WaitForFreeThread` exposes the latter — **any custom `ATaskScheduler` implementation breaks on rebuild.** Accepted deliberately, matching `ISendMessages`, which already carries `Send` and `SendAsync` on one interface.

No changelog entry here: `CHANGELOG.md` has no `Unreleased` section and the 0.11.0 precedent is one pass at release time. The ⚠️ requirement is recorded on #256, which stays open until PR 7 and is what the release notes will be written from.

## The design, and the two things review corrected in it

`ManualResetEventSlim` has no `WaitAsync`, so a `TaskCompletionSource` sits beside it. Every transition — of the reset event, the completion source, and the cancellation token — happens under one lock, because those three must stay coherent. `RunContinuationsAsynchronously` is what makes completing a source inside that lock safe.

Two Critical defects were found and fixed, **both in the design rather than the implementation**:

**The cancellation check was outside the lock**, so the guard wasn't atomic with what it guarded:

```
A: reads the flag -> false          (not yet in the lock)
B: Cancel   -> completes the current source false
C: Reset    -> sees it completed, installs a fresh incomplete source
A: takes the lock, returns that source -- nothing will ever complete it
```

The synchronous `Wait()` returns `false` in that same interleaving, so the two diverged in the hanging direction.

**Making the source lazy then broke `Dispose`.** Laziness was added to keep `Reset` off the scheduler's allocation path (it runs ~twice per message when saturated). But it invalidated the premise that `_asyncWait` is always non-null, so `Dispose` could no-op on a null source while a racing first `WaitAsync` manufactured an incomplete one nothing could complete. Closed by checking `IsDisposed` inside the lock alongside the cancellation check.

Both are latent today — `StopWorkers` always calls `Cancel` before `Dispose` — but they are real holes in the primitive, and PR 2 is when async waiters actually arrive.

## Testing

2,929 tests across all 11 unit projects, zero failures. Release build with `-p:CI=true` clean (`TreatWarningsAsErrors` + XML docs).

Two things worth stating rather than glossing:

- **The `Dispose`-race regression test has unproven detection power.** Roughly 700k racing attempts across several designs — barriers on dedicated threads, CPU oversubscription, swept head-starts — could not force the interleaving; the window is a few instructions. It costs 0.08s and stays as a tripwire, but the fix rests on the code walkthrough and an instrumented harness, not on that test.
- **One test was found to be vacuous and expensive.** `SetAndReset_Concurrently` was 50.08s of the class's 54.4s, and its assertion was almost never evaluated because both probes timed out nearly every iteration. Replaced with a deterministic check that creates the waiter *before* the race and asserts it gets released — a `Set` always runs, so its source must complete whichever order the pair lands in. **Class now runs in 4.39s.**

## Follow-up worth an issue

The 15 pre-existing tests in `WaitForEventOrCancelThreadPoolTests` build their subject through AutoFixture's `AutoNSubstituteCustomization`, which auto-mocks the factory recursively — so they assert "does not throw" and nothing more. The new tests use a real `WaitForEventOrCancel` instead. Not fixed here; out of scope for a no-behaviour-change PR.

## Summary by CodeRabbit

* **New Features**
  * Added non-blocking asynchronous waiting for event completion or cancellation.
  * Added asynchronous waiting for shared and work-group-specific events.
  * Async waits now correctly handle signaling, cancellation, resets, disposal, and multiple concurrent waiters.

* **Tests**
  * Added comprehensive coverage for async wait behavior, race conditions, cancellation, disposal, and work-group isolation.

* **Chores**
  * Elevated ValueTask usage analysis violations to errors.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,6 +26,11 @@ dotnet_diagnostic.CA2263.severity = none
 # CA1859: recommends concrete return types for performance, but the public API
 # deliberately returns interfaces (IQueueOutputMessage, etc.) as the contract.
 dotnet_diagnostic.CA1859.severity = none
+# ValueTask has usage rules a test suite will not catch: a ValueTask over an
+# already-completed result tolerates double-await and .Result silently, and only
+# breaks under IValueTaskSource pooling. The analyzer is deterministic where a test
+# is incidental. Raised to error deliberately - see the async receive path spec.
+dotnet_diagnostic.CA2012.severity = error
 
 # Test projects only: rules that are noise in tests but kept active in production code.
 [**/*Tests*/**.cs]
--- a/Source/DotNetWorkQueue.Tests/Queue/WaitForEventOrCancelTests.cs
+++ b/Source/DotNetWorkQueue.Tests/Queue/WaitForEventOrCancelTests.cs
@@ -139,6 +139,166 @@ namespace DotNetWorkQueue.Tests.Queue
             }
         }
 
+        [TestMethod]
+        public async Task WaitAsync_ReturnsImmediately_WhenAlreadySignaled()
+        {
+            using var test = Create();
+            // constructed signaled, per ManualResetEventSlim(true)
+            Assert.IsTrue(await test.WaitAsync());
+        }
+
+        [TestMethod]
+        public async Task WaitAsync_Blocks_UntilSet()
+        {
+            using var test = Create();
+            test.Reset();
+
+            var waiter = test.WaitAsync().AsTask();
+            Assert.IsFalse(waiter.IsCompleted, "must not complete while reset");
+
+            test.Set();
+            Assert.IsTrue(await waiter);
+        }
+
+        [TestMethod]
+        public async Task WaitAsync_ReturnsFalse_WhenCancelled()
+        {
+            using var test = Create();
+            test.Reset();
+
+            var waiter = test.WaitAsync().AsTask();
+            test.Cancel();
+
+            Assert.IsFalse(await waiter);
+        }
+
+        [TestMethod]
+        public async Task WaitAsync_ResetDuringPendingWait_DoesNotOrphanTheWaiter()
+        {
+            //the lost wake-up case: Reset must not swap out a TCS that already has waiters,
+            //or this waiter sleeps until some later Set that it was never told about
+            using var test = Create();
+            test.Reset();
+
+            var waiter = test.WaitAsync().AsTask();
+            test.Reset();                       // second reset, waiter already pending
+            test.Set();
+
+            var completed = await Task.WhenAny(waiter, Task.Delay(TimeSpan.FromSeconds(5)));
+            Assert.AreSame(waiter, completed, "waiter was orphaned by Reset");
+            Assert.IsTrue(await waiter);
+        }
+
+        [TestMethod]
+        public async Task WaitAsync_SetThenReset_BeforeWaiterResumes_StillReleases()
+        {
+            using var test = Create();
+            test.Reset();
+
+            var waiter = test.WaitAsync().AsTask();
+            test.Set();
+            test.Reset();                       // immediately re-armed
+
+            var completed = await Task.WhenAny(waiter, Task.Delay(TimeSpan.FromSeconds(5)));
+            Assert.AreSame(waiter, completed, "a Set that happened must release its waiter");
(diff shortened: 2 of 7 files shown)


blehnen:
Three findings. I checked each against the code and against the runtime rather than reasoning from the diff — I have already been wrong three times on this particular state machine today, so claims about it get tested. All three are declined, with evidence.

## 3 (High) — "cancelled waits can report success". False positive: it compares two different scenarios

The finding pairs an **already-pending async waiter** (which gets `true`) against a **fresh synchronous `Wait()` call** made afterwards (which returns `false`). Those aren't the same situation, and each path behaves identically to the other when you compare like with like.

**Pending waiter, in exactly the described race** — `Set` wins, `Cancel` arrives immediately after. Run against the real `ManualResetEventSlim`:

```
pending SYNC waiter  -> TRUE (released by Set)
pending ASYNC waiter -> TRUE (TrySetResult(true) landed first; the later TrySetResult(false) is a no-op)
```

Both report success. A blocked `MRES.Wait(token)` that observes the `Set` returns `true` and a subsequent `Cancel` cannot retract it — precisely as `TrySetResult(false)` cannot overwrite a completed source. The sync path has the same race, and resolves it the same way.

**Fresh call after both have happened:**

```
ManualResetEventSlim(true) + already-cancelled token
  -> threw OperationCanceledException -> sync reports FALSE
```

and `WaitAsync` returns `false` from its in-lock `IsCancellationRequested` check. Agreement again.

So there is no ordering where the two disagree. The remedy the finding implies — moving `_cancellationTokenSource.Cancel()` inside `_asyncSync` — would also be actively harmful: `CancellationTokenSource.Cancel` runs registered callbacks **synchronously and inline**, so holding the lock across arbitrary user callbacks trades a non-existent divergence for a real deadlock risk. Both the whole-branch review and the scoped re-review called that out independently.

Worth noting the diff already contains a test for the neighbouring real case (`WaitAsync_AfterCancelThenReset_StillReturnsFalse`), and the genuine defect in this area — the cancellation check sitting *outside* the lock — was found and fixed earlier in this branch at `eb7a36af`.

## 1 and 2 (Medium) — LGPL header on modified test files. Not the convention here

Both files are **pre-existing** and neither carried a header before this PR; it adds test methods to them rather than creating them.

Measured: of 40 sampled files under `Source/DotNetWorkQueue.Tests`, **9 carry the header and 31 do not**. `CLAUDE.md` scopes the requirement to library and production sources, and the license-header sweep (#206) deliberately excluded test projects.

Adding headers to these two would make them inconsistent with the ~78% of sibling test files that have none, and would be unrelated churn in a PR whose point is that it changes no behaviour. Declining on convention grounds, not because the rule is wrong.

If the intent is that *all* test files should carry it, that's a worthwhile repo-wide cleanup — but it belongs in its own change covering all 31, not smuggled into two files because they happened to be touched here.

PR #280, opened 2026-09-08
Plan: PR 2 of the async receive path — IReceiveMessages and the gate

blehnen:
Implementation plan for PR 2 of #256. No code.

## Scope: it stops at the gate

Four tasks, ending with the acceptance test. The spec calls PR 2 the go/no-go point — *"the one to stop at if the starvation twin will not go green"* — so the plan reaches that test in **four** tasks rather than seven:

| task | |
|---|---|
| 1 | `ReceiveMessageAsync` on the interface and its four decorators; all six transports get a temporary sync-inside-async body so the tree compiles |
| 2 | the async consumer awaits the de-queue and the throttle PR 1 made awaitable |
| 3 | Redis made genuinely async |
| 4 | **the gate** — the async twin of `StarvationBaselineTests` |

Tasks 5–7 (SQL Server, PostgreSQL, Memory) are planned **after** the gate passes, and **the PR is not opened until they are done** — so no half-async receive path reaches a consumer.

## Surveying the tree changed task 3

Three things already exist that the spec assumed would need building:

- **`BaseLua.TryExecuteAsync`** — already uses `db.ScriptEvaluateAsync`
- **`DequeueLua.ExecuteAsync(long unixTime)`** — already returns `Task<RedisValue[]>`
- **`IQueryHandlerAsync<TQuery, TResult>`** — already in `Transport.Shared`

The async Lua path is there for the send side. So Redis needs a handler that calls what exists, plus an awaitable work-sub — not a new Lua layer.

And the async dequeue handler differs from the synchronous one by **exactly one line**, which the plan names rather than saying "copy its structure":

```csharp
RedisValue[] result = _dequeueLua.Execute(unixTimestamp);
// becomes
RedisValue[] result = await _dequeueLua.ExecuteAsync(unixTimestamp).ConfigureAwait(false);
```

The other half is `RedisQueueWorkSub.Wait()`, which blocks on its own `ManualResetEventSlim` — so it needs the same `TaskCompletionSource`-beside-`ManualResetEventSlim` treatment PR 1 applied to `WaitForEventOrCancel`. The plan points at that file as the reference and names the four things review corrected there, because getting them wrong yields a lost wake-up rather than a failure.

## What PR 1 taught, carried forward

**The invariants are stated up front.** Receive is not a pure fetch: it attaches `context.Commit` / `Rollback` / `Cleanup`, calls `SetMessageAndHeaders` on three paths, loops over expired messages and resets the work signal. An async rewrite that returns a message and nothing else compiles, returns messages, and passes all four `ConsumerAsync*` scenarios with a dropped cleanup handler.

**Every step says to run builds in the foreground.** Three of six agents stalled on background jobs during PR 1, each time with the work already finished.

**Task 2 has an explicit BLOCKED instruction.** Making `SchedulerMessageHandler` async may require changing a public signature; that is a design decision, not an implementation one, so the plan says to stop and report rather than decide.

## What to review

Mainly the task ordering. Reaching the gate early is the whole design of this plan, but it means tasks 1 and 2 land a consumer that awaits a receive which is still synchronous underneath — correct and behaviour-neutral, but worth agreeing to deliberately rather than discovering.

Also worth a look: task 4 says that if the async starvation test is red, **stop and report BLOCKED** rather than attempt fixes. That is the plan refusing to let an implementer debug its way past the gate.

## Summary by CodeRabbit

- **New Features**
  - Added asynchronous message receiving to improve responsiveness and resource efficiency.
  - Redis message retrieval now supports a fully asynchronous processing path.
  - Async consumers can await message delivery and throttling without blocking worker threads.
  - Added asynchronous receive support across supported message-processing integrations and transports.

- **Tests**
  - Added coverage for asynchronous receiving under thread-starvation conditions.
  - Separated synchronous and asynchronous starvation checks for clearer validation results.

### DIFF
--- a/docs/superpowers/plans/2026-09-09-async-receive-path-pr2.md
+++ b/docs/superpowers/plans/2026-09-09-async-receive-path-pr2.md
@@ -1,0 +1,495 @@
+# Async receive path — PR 2: `IReceiveMessages`
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Give `IReceiveMessages` an async twin, make the async consumer await it instead of blocking a pool thread, and prove the result with the test that currently documents the defect.
+
+**Architecture:** The interface gains `ReceiveMessageAsync`; the four existing decorators each gain a mirroring method (no DI changes, since the interface is the same one). The consumer loop awaits both the receive and the throttle PR 1 made awaitable. Redis is then made genuinely async, because it is the transport the acceptance test exercises.
+
+**Tech Stack:** .NET 10 / .NET 8, MSTest 4.2.3, NSubstitute, StackExchange.Redis 3.0.7, Polly 8.
+
+**Spec:** `docs/superpowers/specs/2026-09-08-async-receive-path-design.md`
+
+## Scope: this plan stops at the gate
+
+The spec calls PR 2 the go/no-go gate — *"the one that matters and the one to stop at if the starvation twin will not go green."* This plan therefore covers **tasks 1 to 4**, ending with that test. Tasks 5 to 7 — SQL Server, PostgreSQL and Memory made genuinely async — are planned separately **after** the gate passes, and the PR is not opened until they are done, so no half-async receive path reaches a consumer.
+
+Reaching the gate in four tasks rather than seven is the point. If the async receive cannot beat a capped thread pool, the design is wrong and we will have spent four tasks finding out.
+
+## What already exists — do not rebuild it
+
+Verified in the tree before writing this plan:
+
+| | |
+|---|---|
+| `BaseLua.TryExecuteAsync` | exists, uses `db.ScriptEvaluateAsync` |
+| `DequeueLua.ExecuteAsync(long unixTime)` | exists, returns `Task<RedisValue[]>` |
+| `IQueryHandlerAsync<TQuery, TResult>` | exists in `Transport.Shared`, used by Dashboard queries |
+| `IWaitForEventOrCancel.WaitAsync()` | added by PR 1 |
+| `IWaitForEventOrCancelThreadPool.WaitAsync(IWorkGroup)` | added by PR 1 |
+| `StarvationBaselineTests` | exists, deliberately RED, excluded via `TestCategory!=StarvationBaseline` |
+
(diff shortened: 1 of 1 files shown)


blehnen:
All three valid, fixed in `1821249c`. The Major one was a genuine hole in the gate, and chasing it turned up a second problem the review hadn't named.

## The gate had no executable oracle — and it would have been excluded from CI

Verified: `StarvationBaselineTests` ends with

```csharp
Assert.IsNull(caughtException, $"Thread-pool starvation reproduced ...");
```

Starvation reproduces, so `caughtException` is non-null and the assertion **fails**. It is a red diagnostic, not a test that passes by expecting an exception. So running it alongside the async twin is exactly as ambiguous as you say: `dotnet test` exits non-zero either way, and one run cannot separate "the async path regressed" from "the baseline is red as intended".

They now run under separate filters with stated expected exit codes — baseline non-zero, gate zero — and the gate passes only on that pair. Plus a case the plan previously had no answer for: **if the baseline starts passing, stop.** That means the pool cap is no longer reproducing the condition the gate is measured against, so a green async test would be meaningless rather than good news.

**Chasing that turned up something worse.** The plan told the implementer to give the async twin `[TestCategory("StarvationBaseline")]`. The Jenkinsfile runs Redis with `--filter "TestCategory!=StarvationBaseline"` — so the new test would have been **excluded from CI permanently**, which is the exact opposite of the spec's plan to promote it to a gate. It now gets its own `StarvationAsync` category, which is what makes it selectable on its own.

## The metrics timer — my instruction was wrong, not under-specified

Correct that the method body and the prose disagreed. The resolution is to delete the prose: the **synchronous** `ReceiveMessage` in that decorator times nothing either — the class's `_waitTimer` belongs to a different member — so adding an async timer would have made the two paths report different metrics for the same operation. Instruction removed rather than implemented.

## Cancellation on the Redis path

Also correct, and the honest fix is documentation rather than plumbing. Nothing downstream accepts a token: `IQueryHandlerAsync.HandleAsync(query)` takes none, `DequeueLua.ExecuteAsync(long unixTime)` takes none, and SE.Redis's `ScriptEvaluateAsync` is not given one.

Threading tokens through those signatures to make it *look* wired would be worse than leaving it. The plan now states this outright and names the two mechanisms that must survive instead — the loop's two `_cancelWork.AnyCancellationRequested()` checks, and the work-sub's own token inside `WaitAsync` — with a note that the parameter still earns its place because SQL Server and PostgreSQL do take real tokens on their async ADO calls in tasks 5 and 6.

Issue #286, opened 2026-09-09
Relational ADO abstraction is built on IDbConnection, which cannot express async


blehnen:
## Problem

The relational transport abstraction is built on `System.Data.IDbConnection` and `IDbCommand`:

- `IDbConnectionFactory.Create()` returns `IDbConnection`
- `ITransactionFactory.Create(IDbConnection)` returns a wrapper over `IDbTransaction`
- command handlers work with the `IDbCommand` that `IDbConnection.CreateCommand()` gives them

Those interfaces date from .NET 1.x and carry **no asynchronous members**. `OpenAsync`, `ExecuteNonQueryAsync` and `CommitAsync` exist only on the `DbConnection`, `DbCommand` and `DbTransaction` base classes, which the abstraction does not expose. `System.Data.Common` is what provider-agnostic code is meant to target now; `IDbConnection` is effectively vestigial.

The practical consequence: **no relational command handler can be made genuinely asynchronous** while this abstraction stands. Discovered while starting #284, where it blocks the async conversion of the commit, rollback, remove, history and poison-message paths on SQL Server, PostgreSQL and SQLite. Redis is unaffected — it goes through Lua on `BaseLua`, which already has `TryExecuteAsync`.

## Approach

Move the abstraction to the `System.Data.Common` base classes. Enabling only — nothing becomes asynchronous in this change, so it can be reviewed as the mechanical refactor it is.

Early sizing, which the compiler will correct:

- `IDbConnectionFactory` and `ITransactionFactory` — the signatures themselves
- the `where TConnection : class, IDbConnection` constraint, which appears in three files
- `SqLiteMessageQueueSharedInit`, the only place instantiating with the raw `<IDbConnection, IDbTransaction, IDbCommand>`; SQLite's real connection is `SQLiteConnection`, so this is a naming laziness rather than an obstacle
- roughly 108 files reference the two factories; how many declare `IDbConnection` locals that need widening is not yet known

Every provider in use already derives from the right base types — `SqlConnection`, `NpgsqlConnection` and `SQLiteConnection` are all `DbConnection`. That is what makes this tractable.

## Breaking

Yes, for anyone implementing `IDbConnectionFactory` or the related abstractions outside this repository. Accepted deliberately: the alternatives are to cast at every call site and degrade silently for providers that do not derive from `DbConnection`, or to leave three transports permanently unable to offer an async path. Neither is correct, only cheaper.

## Blocks

#284

Issue #284, opened 2026-09-09
Remaining synchronous calls on the async receive path — blocks the next release


blehnen:
## Release blocker

This must be resolved before the next release. #282 makes the receive path asynchronous but leaves several synchronous calls on it, and shipping a half-async path is worse than shipping either a fully sync or a fully async one — a consumer advertised as non-blocking that still parks a thread-pool thread on a commit is harder to reason about than one that never claimed otherwise.

## What #282 fixed

- `IReceiveMessages.ReceiveMessageAsync`, awaited by the consumer
- `IQueueWait.WaitAsync`, so the idle and fatal-error back-offs no longer block
- Redis de-queue and work-signal wait, both genuinely asynchronous

## What still blocks, and where it now runs

All of these are sync-only interfaces, verified rather than assumed — each has zero `Task`/`ValueTask` members today:

| Call | Interface | Where it runs after #282 |
|---|---|---|
| `Commit` | `ICommitMessage` | `ProcessMessageAsync.HandleAsync`, **after** awaiting the user's handler — so on a thread-pool continuation |
| `Rollback` | `IRollbackMessage` | `MessageProcessingAsync` catch paths, same continuation |
| `Handle` | `IReceivePoisonMessage` | `TryProcessIncomingMessageAsync` catch path |
| `RecordProcessingStart` and siblings | `IWriteMessageHistory` | inline in `ReceiveMessagesHistoryDecorator.ReceiveMessageAsync`, so on the receive path itself when history tracking is on |
| `Remove` | `IRemoveMessage` | `ReceiveMessageQueryHandlerAsync`, expired-message branch |
| `Send` | `ISendHeartBeat` | `HeartBeatWorker`, via `SmartThreadPoolTaskScheduler` — already on the pool before #282 |

The `Commit` and `Rollback` rows changed character in #282: they used to run on the consumer's dedicated `LongRunning` thread, where blocking costs nothing. Now they run in a continuation after an awaited receive, which is a pool thread.

## Do not convert these on principle — measure first

#282's review turned up a case where making the receive asynchronous *moved* a blocking call from a dedicated worker thread onto a pool thread and made that spot worse, not better. The same trap applies to every row above.

For each, establish before changing anything:

1. Which thread actually runs it — dedicated worker, pool continuation, or scheduler thread. Consumers run their loop under `TaskCreationOptions.LongRunning`, so blocking there is free; blocking a pool thread is not.
[...]

Issue #283, opened 2026-09-09
Re-evaluate the LiteDb transport when LiteDB v6 ships (true async, and the upgrade itself)


blehnen:
## Trigger

LiteDB v6 is expected to add genuine `*Async` methods. That invalidates the *reason* recorded in the LiteDb transport today, so this should be revisited when v6 ships as a stable release.

## Current state

`LiteDbQueueReceiveMessages.ReceiveMessageAsync` (added in #256) returns the synchronous result wrapped in a completed `ValueTask`, with a comment explaining that this is the final implementation rather than a placeholder — because LiteDB 5.0.21 has no async API at all.

That comment is correct for 5.0.21 and will read as a confident, wrong claim once v6 exists. Whoever picks this up should fix the comment even if they change no code.

## What to evaluate, in order

1. **Does it actually help?** Do not assume it does. The reasoning behind the current implementation does not obviously change with v6:
   - Consumers run their loop on a dedicated thread (`TaskCreationOptions.LongRunning`), so a blocking receive holds no thread-pool thread
   - A LiteDb de-queue is on the order of microseconds after #234
   - LiteDB is an embedded file database, so async there means not blocking on file I/O — materially different from the StackExchange.Redis problem in #161, where a synchronous call needs a pool thread to run the completion satisfying its own blocking wait, and deadlocks under pressure

   Measure before changing. The gate built in #256 (`StarvationReceiveBaselineTests`) is a starting shape, though it will need adapting — it is written against Redis and its thresholds are calibrated to reproduce that defect.

2. **The upgrade itself.** A major version bump wants its own assessment: breaking API changes, file format compatibility for existing queues, and whether the performance work in #234 (index-scan removal on the de-queue path, real batch send, dropping the process-wide send lock) still holds or is superseded by v6's engine.

3. **Only then, the async receive**, if 1 shows a benefit.

## Rejected: LiteDB.Async

The third-party `LiteDB.Async` wrapper was considered and declined. It adds a dependency to a published library, with the licensing and maintenance risk that carries, to work around a missing API the vendor is adding natively. Not worth it for a benefit that has not been demonstrated.

## Blocked until

v6 ships as a **stable** release — this project does not pin preview, rc or beta packages.

blehnen:
Folded into #320.

[...]

PR #289, opened 2026-09-09
feat: move a poison message to the error queue without blocking (#284)

blehnen:
Second of the six synchronous calls #284 tracks on the asynchronous receive path. Moving a poison message to the error queue is transport I/O, and the async consumer was calling the blocking member for it.

## Why now

Before #282 this cost nothing. `MessageProcessingAsync` ran under `async void Handle()` on a worker whose thread is dedicated (`TaskCreationOptions.LongRunning`), so blocking there never touched the pool. Once the receive became awaitable, everything after the first `await` runs in a continuation **on a thread-pool thread** — and this catch block is downstream of it. On Redis it is also a sync-over-async completion, which is #161's contention.

It is also the first of the six that the relational transports can answer at all: `IDbConnection` has no `OpenAsync`, so this needed #287 first, and `MoveRecordToErrorQueueCommand` has no output, so it needed #285's `ICommandHandlerAsync<T>`.

## What to look at

**`IReceivePoisonMessage` gains `HandleAsync`** — implemented by all four handlers and the three decorators over it. `MessageProcessingAsync` awaits it; `MessageProcessing` (the synchronous consumer) is unchanged.

**Two transports gain nothing and say so.** The shared `ReceivePoisonMessage<T>` now takes both handlers, so every transport needs the async one:

- **Memory** returns a completed task around the synchronous work. Nothing to release a thread for, and `Task.Run` would only move it to another pool thread.
- **LiteDB** has no async API until v6 (#283), so it runs the synchronous handler on the pool — the hack `SendMessageCommandHandlerAsync` already uses. It frees the caller's thread; it does not make the work asynchronous.

**The relational handler is genuinely async** — `OpenAsync`, `ExecuteNonQueryAsync`, `CommitAsync`, and the three delegated handlers it awaits (`DeleteMetaData`, `SetStatusTableStatus`, `SetStatusTableStatusTransaction`) are async twins rather than wrappers. Both branches of the sync handler are mirrored, including the held-transaction one.

**`ErrorLua.ExecuteAsync` deliberately omits the `Connection.IsDisposed` guard** that `DequeueLua` has, matching its own synchronous twin rather than a different script.

**The registration tests are the ones worth reading.** An `ICommandHandlerAsync` that resolved *without* its retry decorator would move a message to the error queue on the first transient failure instead of retrying — and nothing else in the suite would notice, because the synchronous twin beside it retries either way.

## Breaking

Yes, for anyone implementing `IReceivePoisonMessage` outside this repository. Noted in the changelog. The changelog also picks up the `System.Data.Common` break from #286/#287, which merged without an entry.

## Validation

- Unit suites, 2,398 passed / 0 failed: core 1125, RelationalDatabase 251, SQLite 236, SqlServer 198, Redis 192, PostgreSQL 183, LiteDb 180, Memory 33
- **SQLite integration: 200 passed / 0 failed** — the transport that exercises the new relational handler end to end locally
- New tests fail as expected when the change is reverted — the pacing assertions, and the registration tests with the closed-generic registration removed

Redis, SQL Server and PostgreSQL integration are Jenkins' to confirm; `ConsumerAsyncPoisonMessage` already covers this path on every transport.

## Summary by CodeRabbit

* **New Features**
  * Poison messages are now handled asynchronously, allowing processing to await error-queue transfers without blocking threads.
  * Error handling, logging, tracing, and metrics support the asynchronous poison-message workflow.
  * LiteDB, Redis, relational database, SQLite, and SQL Server transports support asynchronous error-queue transfers.
  * Database transactions can now begin asynchronously.
  * Custom relational transports now use modern database connection and transaction abstractions.

* **Tests**
  * Added coverage for asynchronous poison-message processing, transaction handling, completion ordering, notifications, and transport registration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 - ⚠️ Custom transports and consumers must add async members: `IReceiveMessages.ReceiveMessageAsync`, `IQueueWait.WaitAsync`, and `IRedisQueueWorkSub.WaitAsync` for Redis. Built-in transports are unaffected (GitHub #256)
 - ⚠️ `IMessageProcessing.Handle()` is now `HandleAsync()` and returns a task. Only affects code implementing that interface directly (GitHub #256)
 - Tracing: the async consumer emits its receive span as `ReceiveMessageAsync` rather than `ReceiveMessage`. Dashboards filtering on the old name will not match it. Metric names are unchanged (GitHub #256)
-- The async consumer moves a poison message to the error queue without holding a thread while it waits on the transport (GitHub #284)
-- ⚠️ `IReceivePoisonMessage` gains `HandleAsync`. Only affects code implementing that interface directly (GitHub #284)
+- Async consumers keep working under thread-pool pressure when a message turns out to be poison; moving it to the error queue used to occupy a pool thread for the whole round trip (GitHub #284)
+- ⚠️ `IReceivePoisonMessage` gains `HandleAsync` and `ITransactionWrapper` gains `BeginTransactionAsync`. Only affects code implementing those interfaces directly (GitHub #284)
 - ⚠️ Custom relational transports: `IDbConnectionFactory`, `ITransactionFactory` and `ITransactionWrapper` now use `System.Data.Common` types (`DbConnection`, `DbTransaction`) rather than the `System.Data` interfaces, which have no async members. Built-in transports are unaffected (GitHub #286)
 
 ### 0.11.0 — 2026-09-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 - ⚠️ Custom transports and consumers must add async members: `IReceiveMessages.ReceiveMessageAsync`, `IQueueWait.WaitAsync`, and `IRedisQueueWorkSub.WaitAsync` for Redis. Built-in transports are unaffected (GitHub #256)
 - ⚠️ `IMessageProcessing.Handle()` is now `HandleAsync()` and returns a task. Only affects code implementing that interface directly (GitHub #256)
 - Tracing: the async consumer emits its receive span as `ReceiveMessageAsync` rather than `ReceiveMessage`. Dashboards filtering on the old name will not match it. Metric names are unchanged (GitHub #256)
+- The async consumer moves a poison message to the error queue without holding a thread while it waits on the transport (GitHub #284)
+- ⚠️ `IReceivePoisonMessage` gains `HandleAsync`. Only affects code implementing that interface directly (GitHub #284)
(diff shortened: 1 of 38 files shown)


blehnen:
Review round pushed in 29fa49c0. Two of the findings were real behaviour bugs, and both are worth a second look in review.

## Fixed

**SQLite lost its database-existence guard on the async path** (both reviewers, High). The synchronous handler is decorated with `MoveRecordToErrorQueueCommandDecorator`, which declines the move when the file is gone. The async registration had no equivalent, so a deleted queue turned a poison message into a command failure and a retry cycle instead of a decline. SQLite is the only transport whose store can simply disappear, which is why only it carries that decorator. Added `MoveRecordToErrorQueueCommandDecoratorAsync` plus registration and three tests, checked to fail with the guard removed.

**`BeginTransaction()` was still synchronous** right after `await conn.OpenAsync()`. On SQL Server and PostgreSQL that is a round trip, so the non-blocking path blocked on it — and both async send paths already await `BeginTransactionAsync` directly, so the providers plainly support it. `ITransactionWrapper` gains `BeginTransactionAsync`, implemented in the relational wrapper, the SQLite wrapper and SQLite's Polly retry decorator, whose pipeline lookup is now shared between the two members rather than duplicated. Covered by three new decorator tests; SQLite integration re-run at 200/200 since this touches its transaction path.

**Tracing never set the activity status to Error in the no-header branch.** Fixed in the synchronous branch too — the async one was written as a faithful twin, so it inherited the defect. That is a useful signal about this whole conversion: mirroring the synchronous path exactly also mirrors its bugs.

**`ErrorLua` doc ordering** — `ExecuteAsync` had been inserted between `GetParameters`' doc comment and the method, orphaning it. **LiteDb registration test** now uses a unique temp path per run with cleanup, rather than one fixed name.

**Changelog** entry reworded to carry a before/after effect pair.

## Not changed

**License headers on the new test files.** Presented as the project standard, but the test projects are the opposite: 19/191 files in `DotNetWorkQueue.Tests` carry one, 1/19 in `Transport.Memory.Tests`, and the newest test file in the repo (added by #287) has none. #206 scoped the header sweep to non-test production files deliberately. Adding headers to eight new files would make those the outliers.

**`BaseLua.TryExecuteAsync`'s `NOSCRIPT` recovery calls the synchronous `LoadScript()`.** Correct finding, but pre-existing and shared by every async Lua caller including the send and de-queue paths — not this PR's to fix, and worth measuring before assuming an async loader is a visible win. Filed as #290.

**"The new transport handlers lack tests."** `ConsumerAsyncPoisonMessage` already covers this path end to end on every transport; six of those ran locally on SQLite as part of 200/200. The claim that the new handlers sit outside end-to-end coverage is not accurate.

## Validation

2,404 unit tests / 0 failed. SQLite integration 200 / 0 failed. `NoTests.sln -c Release -p:CI=true` clean.

Issue #290, opened 2026-09-09
Redis: NOSCRIPT recovery on the async Lua path loads the script synchronously


blehnen:
`BaseLua.TryExecuteAsync` awaits `ScriptEvaluateAsync`, but its `NOSCRIPT` recovery calls the synchronous `LoadScript()`:

```csharp
catch (RedisException e)
{
    if (e.Message.StartsWith("NOSCRIPT", StringComparison.InvariantCultureIgnoreCase))
    {
        LoadScript();   // <-- synchronous
        ...
    }
}
```

`LoadScript()` enumerates the endpoints and calls `luaScript.Load(server)` against **every** server, so a Redis restart, a failover, or a `SCRIPT FLUSH` makes the recovery path block its continuation thread for the duration — on a thread-pool thread, which is what the async path exists to avoid.

## Scope

This is not new and it is not specific to one script. It affects every caller of `TryExecuteAsync`, so it already applies to the async send and de-queue paths, and now also to the poison-message move added in #289.

## What to do

- Add an asynchronous load using StackExchange.Redis' `ScriptLoadAsync`, awaited from `TryExecuteAsync` during `NOSCRIPT` recovery
- Keep the existing synchronous loader for `TryExecute`
- Consider serialising or sharing concurrent reloads, so N simultaneous cache misses do not each reload every endpoint

## Notes

Only reachable when Redis reports `NOSCRIPT`, so it is a recovery-path cost rather than a steady-state one. Worth measuring the reload cost against a multi-endpoint setup before assuming the async version is a visible win — the value here is not blocking a pool thread, not throughput.

Found by review on #289. Related: #161 (SE.Redis sync-over-async contention), #284.

blehnen:
Fixed by #305.

Issue #167, opened 2026-06-29
Inbox held-transaction batch send: true multi-row insert in the caller's transaction (SQL Server + PostgreSQL)


blehnen:
## Summary

Follow-up to #162. The true bulk-insert batch send shipped for the standalone/outbox path, but the **inbox held-transaction** batch overloads (`EnableHoldTransactionUntilMessageCommitted`) still loop single-row inserts. This issue tracks giving that path a true multi-row insert inside the caller's transaction.

This was explicitly deferred during #162 phase 1: the standalone batch path shipped, and the inbox shared-transaction batching was split out to here.

## Current behavior

`SqlServerRelationalProducerQueue.SendWithExternalTransactionBatch` / `...BatchAsync` validate the caller's `DbTransaction` once, then loop `SendOne` per message (a `RelationalSendMessageCommand`, single-row insert) inside that transaction. This is correct and atomic, because the caller's transaction provides the atomicity, but it is still N single-row round trips. There is no multi-row batching on the inbox path.

The base `RelationalProducerQueue.SendWithExternalTransactionBatch[Async]` throws `InvalidOperationException` by default; transports opt in by overriding.

## What already exists

- `RelationalSendMessageCommandBatch` carries an optional `DbTransaction ExternalTransaction` and implements `IRetrySkippable` (`SkipRetry` is true when an external transaction is present, so the retry decorator stays out of the caller's transaction).
- The override hook points (`SendWithExternalTransactionBatch[Async]`) are in place on `SqlServerRelationalProducerQueue`.

## What needs doing

1. The batch command handler (`SendMessageCommandBatchHandler` / `...Async`) currently opens its own connection and transaction and commits at the end. When `RelationalSendMessageCommandBatch.ExternalTransaction` is set, it must instead use the caller's connection/transaction for the multi-row insert and meta/status rows, and must not commit (the caller owns the commit).
2. Wire `SendWithExternalTransactionBatch[Async]` to dispatch a `RelationalSendMessageCommandBatch` (with the external transaction) to that handler instead of looping `SendOne`.

## Scope

- SQL Server first, then PostgreSQL once the #162 PostgreSQL batch handler lands.
- SQLite is excluded. The hold-transaction inbox pattern is structurally non-viable on SQLite's single-writer model (holding the transaction blocks the queue's other writers, including the worker's own next dequeue). See #149 (closed); SQLite is outbox-only.

## Benefit and semantics

[...]

PR #291, opened 2026-09-09
feat: an awaitable remove, and the async delete handlers under it (#284)

blehnen:
Third of the six synchronous calls #284 tracks. Depends on #285's `ICommandHandlerAsync` and #287's move to `System.Data.Common`.

## Why `Remove` and not `Commit`

The issue's table lists `Remove` as the expired-message branch of the Redis receive handler. Surveying it first turned up that it is also the per-message commit path:

```
ProcessMessageAsync.HandleAsync
  → CommitMessage.Commit(context)
    → context.RaiseCommit()            // the event, and the remaining Kind-B problem
      → TransportCommitMessage.Commit
        → IRemoveMessage.Remove(context, RemoveMessageReason.Complete)
```

So `Remove` is the leaf `Commit` bottoms out in, not a sibling of it. An awaitable commit event over a blocking `Remove` would buy nothing, which is why this one comes first. Recorded on the issue [here](https://github.com/blehnen/DotNetWorkQueue/issues/284#issuecomment-5608163149).

## Nothing but Redis calls it yet

**Read this before reviewing the relational implementations.** `IRemoveMessage.RemoveAsync` has one production caller in this PR: the Redis expired-message branch. SQL Server, PostgreSQL, SQLite and LiteDB all implement it, and nothing invokes those until the commit event is awaitable in the next PR.

Deliberate, and the same shape as #285. The alternative was one pull request carrying the interface, the leaf handlers, the `IMessageContext` events and every transport's commit together — past the 100-file mark where CodeRabbit's free plan declines to review, which is what happened to #287.

## What to look at

**The Redis receive handler is the only structural change.** Its expiry branch lived inside `AReceiveMessageQueryHandler.BuildMessage`, which the synchronous and asynchronous handlers share on purpose: the parse, the poison detection and the three `SetMessageAndHeaders` calls are load-bearing, and a divergence between two copies would surface on one path only. Removing the message is the one step that cannot be shared — one blocks, the other awaits — so `BuildMessage` now reports the expiry through an `out` parameter and each handler carries it out itself.

Both handlers keep the `try`/`catch` that removal had while it lived inside `BuildMessage`, so a failed removal still surfaces as `ReceiveMessageException` rather than escaping raw. That is asserted, because moving a call out of a `try` block is exactly the kind of change that silently alters what a caller sees.

**The async delete handlers underneath.** Genuinely asynchronous on the relational transports — `OpenAsync`, `BeginTransactionAsync`, `ExecuteNonQueryAsync`, `CommitAsync` — including the held-transaction path. `DeleteLua.ExecuteAsync` for Redis. LiteDB delegates to the thread pool: it has no asynchronous API until v6 (#283), so that frees the caller's thread without making the work asynchronous.

**Two pre-existing defects, fixed on both twins.** The async versions were written by mirroring their synchronous originals, and inherited two bugs from them:

- `DeleteTransactionalMessageCommandHandler` cleared MetaData, Queue, ErrorTracking and optionally Status, but never MetaDataErrors, which the ordinary delete handler does clear. With `EnableHoldTransactionUntilMessageCommitted` on, a message that had errored kept its error rows on SQL Server and PostgreSQL. Dates to `1ffec1e8`.
- `RemoveMessageDecorator` read headers before checking there was an id to read them by. Redis' `GetHeader` dereferences the id unconditionally, so tracing threw on exactly the input every `IRemoveMessage` answers with `NotFound`, making those null guards unreachable whenever tracing was on.

**A new expired-message test for the async consumer.** There was none — only for the synchronous consumer and `ConsumerMethod` — so the behaviour this PR adds reached no test. Redis is the transport it exists for: every other transport expires messages from its own monitor on its own thread, while Redis notices expiry inside the de-queue and removes the message there, which is the only place the removal runs on the asynchronous receive path.

**Two branches this PR cannot unit-test.** The SQL Server and PostgreSQL held-transaction paths in `RemoveAsync` — `SqlTransaction` and `NpgsqlTransaction` are sealed and a connection holder cannot be faked, the same wall #167 hit. Integration covers them once a caller exists, so treat that code as unverified for now.

## Patch coverage is at its ceiling here

Codecov reports the patch around 54%, and it cannot go higher in this PR. Every remaining uncovered line is unreachable rather than untested:

| | lines |
|---|---|
| SQL Server + PostgreSQL `RemoveAsync` | 88 |
| Relational `DeleteMessageCommandHandlerAsync` | 26 |
| `DeleteStatusTableStatusCommandHandlerAsync` | 11 |
| LiteDB `DeleteMessageCommandHandlerAsync` | ~4 |

The delete handlers are injected only into the `RemoveMessage` implementations, so they are reachable only through `RemoveAsync` — which nothing calls on those four transports yet. The Redis chain, which does have a caller, is covered by the new expired-message test.

Giving those lines a caller is the commit/rollback PR, not this one.

## Breaking

`IRemoveMessage` gains both `RemoveAsync` overloads; `ITransactionWrapper` gained `BeginTransactionAsync` in #289's round. Only affects code implementing those interfaces outside this repository; noted in the changelog.

## Also here

`CLAUDE.md`'s licence-header rule said "all source files"; the test projects have never followed it (19 of 191 files in `DotNetWorkQueue.Tests`, 1 of 19 in `Transport.Memory.Tests`) and #206 scoped its sweep to production files deliberately. The rule now says production files and states that test files are exempt. Every production file does comply — verified.

## Validation

- 2,425 unit tests / 0 failed
- **SQLite integration 203 / 0 failed**, including the three new expired-message cases
- Tests checked for teeth: the Redis expiry tests fail if the asynchronous handler calls the blocking member, the metadata-errors tests fail with that statement removed, and the decorator tests fail with the id guard removed

Filed from review rather than folded in: #290 (`NOSCRIPT` recovery loads Lua scripts synchronously) and #292 (the delete handlers return a hardcoded `1`, so `Remove` reports `Removed` for a message that was not there).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 - Tracing: the async consumer emits its receive span as `ReceiveMessageAsync` rather than `ReceiveMessage`. Dashboards filtering on the old name will not match it. Metric names are unchanged (GitHub #256)
 - Async consumers keep working under thread-pool pressure when a message turns out to be poison; moving it to the error queue used to occupy a pool thread for the whole round trip (GitHub #284)
 - Redis async consumers no longer hold a thread while removing an expired message during a de-queue (GitHub #284)
-- ⚠️ `IReceivePoisonMessage` gains `HandleAsync` and `ITransactionWrapper` gains `BeginTransactionAsync`, `IRemoveMessage` gains `RemoveAsync`. Only affects code implementing those interfaces directly (GitHub #284)
+- Fix: SQL Server and PostgreSQL left a message's error-history rows behind when it was removed with `EnableHoldTransactionUntilMessageCommitted` on (GitHub #284)
+- ⚠️ Custom transports and consumers must add async members: `IReceivePoisonMessage.HandleAsync`, `IRemoveMessage.RemoveAsync`, and `ITransactionWrapper.BeginTransactionAsync` for relational transports. Built-in transports are unaffected (GitHub #284)
 - ⚠️ Custom relational transports: `IDbConnectionFactory`, `ITransactionFactory` and `ITransactionWrapper` now use `System.Data.Common` types (`DbConnection`, `DbTransaction`) rather than the `System.Data` interfaces, which have no async members. Built-in transports are unaffected (GitHub #286)
 
 ### 0.11.0 — 2026-09-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 - ⚠️ `IMessageProcessing.Handle()` is now `HandleAsync()` and returns a task. Only affects code implementing that interface directly (GitHub #256)
 - Tracing: the async consumer emits its receive span as `ReceiveMessageAsync` rather than `ReceiveMessage`. Dashboards filtering on the old name will not match it. Metric names are unchanged (GitHub #256)
 - Async consumers keep working under thread-pool pressure when a message turns out to be poison; moving it to the error queue used to occupy a pool thread for the whole round trip (GitHub #284)
-- ⚠️ `IReceivePoisonMessage` gains `HandleAsync` and `ITransactionWrapper` gains `BeginTransactionAsync`. Only affects code implementing those interfaces directly (GitHub #284)
+- Redis async consumers no longer hold a thread while removing an expired message during a de-queue (GitHub #284)
+- ⚠️ `IReceivePoisonMessage` gains `HandleAsync` and `ITransactionWrapper` gains `BeginTransactionAsync`, `IRemoveMessage` gains `RemoveAsync`. Only affects code implementing those interfaces directly (GitHub #284)
 - ⚠️ Custom relational transports: `IDbConnectionFactory`, `ITransactionFactory` and `ITransactionWrapper` now use `System.Data.Common` types (`DbConnection`, `DbTransaction`) rather than the `System.Data` interfaces, which have no async members. Built-in transports are unaffected (GitHub #286)
 
 ### 0.11.0 — 2026-09-06
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -119,7 +119,7 @@ Projects target net10.0 and net8.0. Legacy conditional compilation symbols (NETF
 
 ## Conventions
 
-- All source files include LGPL-2.1 license headers (see `DotNetWorkQueue.licenseheader`)
+- Library and production source files include LGPL-2.1 license headers (see `DotNetWorkQueue.licenseheader`). Test projects are exempt and mostly do not carry one; #206 scoped its header sweep to production files deliberately. Do not add headers to test files to satisfy this rule, and do not flag their absence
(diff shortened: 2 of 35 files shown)


blehnen:
Review round pushed in b1dda30f and 0e6396e9.

**Both substantive findings were pre-existing defects in the synchronous code that the new asynchronous twins copied faithfully.** Fixed on both sides — fixing only the new half would leave the two paths diverging, which is the thing mirroring was protecting against in the first place.

## Fixed

**The held-transaction delete never removed the metadata-error rows** (flagged by both reviewers). `DeleteTransactionalMessageCommandHandler` clears MetaData, Queue, ErrorTracking and optionally Status; the ordinary `DeleteMessageCommandHandler` clears MetaDataErrors as well. So on SQL Server and PostgreSQL, with `EnableHoldTransactionUntilMessageCommitted` on, a message that had errored before being removed kept its error rows and nothing collects them later. The asymmetry dates to the transport refactor in `1ffec1e8`, where the statement was added to one handler and not the other. Four tests over both twins, checked by reverting the fix. Changelog entry added, since a consumer with that option on has rows in their database because of this.

**The tracing decorator read headers before checking there was an id.** Redis' `GetHeader` dereferences `id.Id.Value` unconditionally, so a null or valueless id threw as soon as tracing was in the chain — on exactly the input every `IRemoveMessage` implementation is written to answer with `NotFound`. Those null guards were unreachable whenever tracing was on, which is why nothing had noticed. Fixed on both members, four tests, also checked by reverting.

**Changelog** reshaped to the form the existing #256 entry already uses: one bullet naming the async members a custom transport must add, rather than one per interface. Kept combined for consistency with that precedent rather than split.

## License headers — resolved by changing the rule

Raised on #289 as well, so it is worth settling rather than answering twice. `CLAUDE.md` said "all source files", and the test projects have never followed it: 19 of 191 files in `DotNetWorkQueue.Tests` carry a header, 1 of 19 in `Transport.Memory.Tests`, and #206 scoped its sweep to production files on purpose.

Since that file is what the review bots read as their coding guideline, the stale wording is the actual cause of the finding. `CLAUDE.md` now scopes the rule to library and production files and says explicitly not to flag test files (`0e6396e9`). Verified while making the change: every production file does comply, so a production file without a header remains a real finding. The three test files in this PR keep the headers they were given — harmless, and not worth another commit to undo.

## Not changed

**The delete handlers return a hardcoded `1` rather than the affected row count**, so `Remove` reports `Removed` for a message that was not there. Correct finding and long-standing on both handlers and both transports. Changing what `Remove` returns is a behaviour change on the synchronous path that reaches `ClearExpiredMessages` and the dashboard delete, so it deserves its own review — filed as #292.

**`BaseLua.TryExecuteAsync`'s `NOSCRIPT` recovery is still synchronous.** Already filed as #290 from the #289 review; pre-existing and shared by every async Lua caller, including the send and de-queue paths.

## Validation

2,425 unit tests / 0 failed. SQLite integration 200 / 0 failed. `NoTests.sln -c Release -p:CI=true` clean.

PR #293, opened 2026-09-10
feat: record the processing-start history row without blocking (#284)

blehnen:
Fourth of the calls #284 tracks. Small and self-contained: one interface member, five implementations, one call site.

## What blocks

`ReceiveMessagesHistoryDecorator` writes the processing-start row inline on the receive path. On the asynchronous consumer that write lands in a continuation on a thread-pool thread — once per message, whenever `EnableHistory` and `TrackProcessing` are on.

`IWriteMessageHistory` gains `RecordProcessingStartAsync`. Genuinely asynchronous on the relational transports (`OpenAsync`, `ExecuteNonQueryAsync`) and on Redis (`HashGetAsync`, `HashSetAsync`). Memory returns a completed task — the history lives in process, so there is nothing to release a thread for. LiteDB delegates to the thread pool, having no asynchronous API until v6 (#283): that frees the caller's thread without making the work asynchronous.

## Only one of the seven members

The other six are reached from the commit, rollback, error and send paths, none of which is awaitable yet. An async twin for each would be code nothing calls, and #291 showed what that does to a coverage report — its patch was capped near 54% by exactly that. They come with their callers.

## What to look at

**The tests in this class were asserting the wrong member.** The existing asynchronous tests checked `RecordProcessingStart` — the blocking one — so after this change they were verifying the wrong thing rather than failing. They now assert the asynchronous member *and* that the blocking one is not used, which is the assertion that would otherwise pass either way. One test is new: the receive must not hand the message on before the history row says processing started.

**A sibling check, not a change.** The six history decorators gate on `EnableHistory` plus a per-event `Track*` flag — except rollback, which has only `EnableHistory`. That one is correct: there is no `TrackRollback` option for it to check. Noted because it looks like an omission next to the others.

## Breaking

`IWriteMessageHistory` gains `RecordProcessingStartAsync`. Only affects code implementing that interface outside this repository; in the changelog.

## Two things this PR deliberately does not do

**HeartBeat is not a mirror job.** `ISendHeartBeat.Send` runs inside `lock (_runningLocker)` and `await` is illegal in a lock; `IHeartBeatScheduler.AddUpdateJob` takes `Expression<Action<...>>`, which cannot contain `await` either; and `Stop()` relies on that same lock to block until an in-flight beat finishes. Converting it means replacing the concurrency guard and re-proving the shutdown protocol — its own change, not a ride along here.

**A seventh synchronous call, absent from the issue's table.** `ProcessMessageAsync.cs:89` calls `MessageExceptionHandler.Handle` synchronously on the async continuation, reaching `IReceiveMessagesError.MessageFailedProcessing` and from there `RecordError` and the transport's error write. It runs through the same handler as rollback, so it belongs with that work.

## Validation

- 2,426 unit tests / 0 failed
- **SQLite integration 203 / 0 failed**
- Both new assertions checked by reverting the decorator to the blocking call

## Summary by CodeRabbit

- **New Features**
  - Added asynchronous processing-start history recording across supported transports.
  - Asynchronous consumers now record processing history without blocking the receiving caller thread where supported.
  - Message delivery waits for processing history to complete before continuing.

- **Breaking Changes**
  - Custom relational transports implementing message history must support `RecordProcessingStartAsync`.

- **Bug Fixes**
  - Improved asynchronous history error propagation and processing-status reliability.
  - Corrected cleanup behavior to ensure test queues and history records are removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Async consumers keep working under thread-pool pressure when a message turns out to be poison; moving it to the error queue used to occupy a pool thread for the whole round trip (GitHub #284)
 - Redis async consumers no longer hold a thread while removing an expired message during a de-queue (GitHub #284)
 - Fix: SQL Server and PostgreSQL left a message's error-history rows behind when it was removed with `EnableHoldTransactionUntilMessageCommitted` on (GitHub #284)
-- Async consumers with history tracking on no longer hold a thread while recording that a message started processing (GitHub #284)
+- SQL Server, PostgreSQL and Redis async consumers no longer hold a thread recording that a message started processing, when history tracking is on (GitHub #284)
 - ⚠️ Custom transports and consumers must add async members: `IReceivePoisonMessage.HandleAsync`, `IRemoveMessage.RemoveAsync`, `IWriteMessageHistory.RecordProcessingStartAsync`, and `ITransactionWrapper.BeginTransactionAsync` for relational transports. Built-in transports are unaffected (GitHub #284)
 - ⚠️ Custom relational transports: `IDbConnectionFactory`, `ITransactionFactory` and `ITransactionWrapper` now use `System.Data.Common` types (`DbConnection`, `DbTransaction`) rather than the `System.Data` interfaces, which have no async members. Built-in transports are unaffected (GitHub #286)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 - Async consumers keep working under thread-pool pressure when a message turns out to be poison; moving it to the error queue used to occupy a pool thread for the whole round trip (GitHub #284)
 - Redis async consumers no longer hold a thread while removing an expired message during a de-queue (GitHub #284)
 - Fix: SQL Server and PostgreSQL left a message's error-history rows behind when it was removed with `EnableHoldTransactionUntilMessageCommitted` on (GitHub #284)
-- ⚠️ Custom transports and consumers must add async members: `IReceivePoisonMessage.HandleAsync`, `IRemoveMessage.RemoveAsync`, and `ITransactionWrapper.BeginTransactionAsync` for relational transports. Built-in transports are unaffected (GitHub #284)
+- Async consumers with history tracking on no longer hold a thread while recording that a message started processing (GitHub #284)
+- ⚠️ Custom transports and consumers must add async members: `IReceivePoisonMessage.HandleAsync`, `IRemoveMessage.RemoveAsync`, `IWriteMessageHistory.RecordProcessingStartAsync`, and `ITransactionWrapper.BeginTransactionAsync` for relational transports. Built-in transports are unaffected (GitHub #284)
 - ⚠️ Custom relational transports: `IDbConnectionFactory`, `ITransactionFactory` and `ITransactionWrapper` now use `System.Data.Common` types (`DbConnection`, `DbTransaction`) rather than the `System.Data` interfaces, which have no async members. Built-in transports are unaffected (GitHub #286)
 
 ### 0.11.0 — 2026-09-06
--- a/Source/DotNetWorkQueue.IntegrationTests.Shared/History/Implementation/SimpleHistoryAsyncTest.cs
+++ b/Source/DotNetWorkQueue.IntegrationTests.Shared/History/Implementation/SimpleHistoryAsyncTest.cs
@@ -133,8 +133,14 @@ namespace DotNetWorkQueue.IntegrationTests.Shared.History.Implementation
                 }
                 finally
                 {
-                    oCreation?.RemoveQueue();
-                    oCreation?.Dispose();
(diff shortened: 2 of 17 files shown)


blehnen:
Review round pushed in f9c57fc0, and the coverage number turned out to be the most useful comment on this PR.

## The 5.71% was a real gap, not an artifact

All five `RecordProcessingStartAsync` implementations read 0.00% — including Memory and the no-op, which run in process. Nothing called the member at all.

History is only ever exercised through `CreateConsumer`, the **synchronous** consumer. No test on any transport ran the asynchronous consumer with history enabled, so the write this branch converts reached nothing. That is the same shape as the expiry gap in #291 and the Redis History case `CLAUDE.md` records.

`SimpleHistoryAsyncTest` mirrors its synchronous twin through `CreateConsumerQueueScheduler`, with wrappers for the six transports that have a history test. Six of those cases run locally (Memory, SQLite, LiteDB) and pass; SQL Server, PostgreSQL and Redis are Jenkins'.

**What the new test proves, and what it does not.** It executes the asynchronous member and shows history is recorded on that consumer, which nothing previously checked. It does *not* distinguish which member the decorator calls — reverting to the blocking one leaves it green, which I verified. The unit test is what pins that, and it does fail on revert.

Also asserted, because nothing asserted them anywhere: `StartedUtc` and `DurationMs`. The latter is derived from the former and silently becomes zero when the start row is missing.

While investigating I had assumed the existing synchronous history tests were blind to the processing-start transition. They are not: `RecordComplete` only matches a record already in `Processing`, so asserting `Complete` proves the start row was written. The synchronous path was covered all along.

## Two claims corrected — both reviewers were right

**LiteDB and SQLite do not get what the changelog claimed.** It said async consumers "no longer hold a thread". LiteDB's `Task.Run` still occupies a pool thread, and `System.Data.SQLite` does not override the inherited asynchronous members, so the base class runs them on the calling thread. The entry now names the three transports where the claim is true: SQL Server, PostgreSQL, Redis.

**The relational implementation now says so at the code**, so the async members are not read as a SQLite win later. It is still written this way because all three transports share that implementation, and a per-provider split would be worse than a member that is merely no faster on one of them.

## Sonar

S2077 on the new method is accepted rather than changed. The table name comes from `TableNameHelper` — internal queue configuration, not user input — and a table name cannot be a SQL parameter. Ten identical interpolations in that same file are already accepted; this is the async twin of one of them.

## Not changed

**Licence header on the modified test file.** `CLAUDE.md` scopes that rule to library and production files and says explicitly not to flag test files, amended in `0e6396e9` after the same comment on #289 and #291. Every production file does comply.

**Splitting the breaking-change bullet per interface.** `CLAUDE.md` asks for one bullet per user-visible change rather than one per aspect, and the existing #256 entry sets the pattern by listing several async members a custom transport must add in a single ⚠️ entry. For a consumer this is one change: implement a custom transport, add these members.

## Validation

2,426 unit tests / 0 failed. Memory, SQLite and LiteDB async history tests pass locally. `NoTests.sln -c Release -p:CI=true` clean.

blehnen:
Pushed in 6e37b091. One new finding, and it is a real one.

## The history tests never removed their queue

Both tests dispose the creation object and null it before `finally` runs, so `oCreation?.RemoveQueue()` found nothing and the queue was never removed. Every other shared test removes it there.

The early disposal is deliberate — it releases the connections before the verification container reads the history back, which LiteDB in particular needs — so the fix re-acquires a creation object in `finally` rather than dropping it. `RemoveQueue` is now called unguarded, so a failure fails the run instead of passing quietly.

The old behaviour left queues and history rows behind. On SQLite and LiteDB that is a temp file which gets cleaned up anyway; on **Redis, SQL Server and PostgreSQL it is residue in a shared instance** that nothing else collects.

Pre-existing in the synchronous test and inherited by the asynchronous one when it was mirrored, so fixed in both. Verified on the three transports that run locally — Memory, SQLite and LiteDB, synchronous and asynchronous.

## Already addressed before these comments landed

Three of the remaining findings were raised against `108117b5` and are fixed in `f9c57fc0`, which landed after them:

- **The LiteDB thread-use claim in the changelog.** The entry now names only the transports where it holds — SQL Server, PostgreSQL and Redis — rather than claiming it generally.
- **SQLite still blocking on history.** Correct, and now stated at the code: `System.Data.SQLite` does not override the inherited asynchronous members, so the base class runs them on the calling thread. The implementation stays shared across the three relational transports on purpose; a per-provider split would be worse than a member that is merely no faster on one of them.

## Not changed

**Licence headers on the test files.** `CLAUDE.md` scopes that rule to library and production files and says explicitly not to flag test files — amended in `0e6396e9` after the same comment on #289 and #291. Every production file does comply, so a production file without one remains a real finding.

**Splitting the breaking-change bullet per interface.** `CLAUDE.md` asks for one bullet per user-visible change rather than one per aspect, and the existing #256 entry sets the pattern by listing several async members a custom transport must add in one ⚠️ entry. For a consumer this is a single change: implement a custom transport, add these members.

Issue #294, opened 2026-09-10
Heartbeat can fail to protect an in-flight message, so the queue delivers it twice


blehnen:
`ConsumerHeartbeat` intermittently reports more processed messages than were sent. Seen twice, on different data rows, on two different branches:

| PR | Row | Expected | Actual |
|---|---|---|---|
| #291 | `(7, 45, 280, 3, inMemoryDb: true, enableChaos: true)` | 7 | 8 |
| #294 (commit conversion) | `(7, 45, 180, 3, inMemoryDb: false, enableChaos: false)` | 7 | 9 |

## Why it is the test, not the change

- **Passes in isolation.** 3/3 both times, including the row that had just failed. It only fails inside the full 200+ test run.
- **The second occurrence is on the non-chaos row**, so injected faults are not required to trigger it.
- **It exercises the synchronous consumer** — `ConsumerCancelWorkShared` uses `creator.CreateConsumer(...)`. The branch it failed on converts only the *asynchronous* consumer's commit; `ProcessMessage.cs` and `MessageProcessing.cs` are untouched in that diff.

## The race

`ConsumerCancelWorkShared.RunConsumerInternal` asserts `messageCount == processedCount`. The scenario is deliberately a race: a "bad" consumer takes messages, is disposed mid-flight, the heartbeat monitor resets its in-flight messages, and a good consumer picks them up. If the monitor resets a message the good consumer is *already* working — or the bad consumer completed one before disposal — the message is processed twice and the count overshoots.

Under full-suite load the workers are slower, which widens the window. The overshoot size tracks that: 8 the first time, 9 the second.

## What to consider

The assertion may simply be wrong for what the scenario can guarantee. The test is about the heartbeat resetting stalled work, and reprocessing is the mechanism it relies on — so an exact-equality count is a stronger claim than the design supports. `IsGreaterThanOrEqualTo(messageCount, ...)` plus a separate assertion that every message reached a terminal state may be what was meant.

Worth deciding deliberately rather than re-running until green, which is what happened on #291.

## Related

- Note that `ConsumerHeartbeat` and `ConsumerCancelWork` are also the two scenarios with **no** asynchronous-consumer equivalent (audited 2026-09-10), so whatever is decided here should shape the `ConsumerAsyncHeartbeat` test that #284's heartbeat conversion needs.

blehnen:
This is a bug; it never happened until our async changes. Needed to be addressed before the next version can ship

[...]

PR #295, opened 2026-09-10
feat: commit a message without blocking on the async consumer (#284)

blehnen:
The busiest of the calls #284 tracks. Committing runs on every message, on the happy path, and on the asynchronous consumer it runs in a continuation on a thread-pool thread — so the synchronous commit held one of those for a transport round trip per message.

Rollback and the error path follow in a second pull request; see the end.

## The obstacle was the event, not the commit

`IMessageContext.Commit` is an `EventHandler`, which returns void. The transports subscribe to that event and commit from it, so a subscriber doing transport I/O had no way to be awaited.

This adds `AsyncEventHandler` — a `Task`-returning delegate — and a `CommitAsync` event beside the existing one, with `RaiseCommitAsync` awaiting each subscriber.

**Subscribers are awaited sequentially, in subscription order, not through `WhenAll`.** That matches what the synchronous event already did, and it is load-bearing rather than incidental: the transports commit against a connection and transaction held in the message context, which cannot take concurrent work. Asserted directly, because `WhenAll` would look like an obvious improvement to a later reader.

## What to look at

**Each receive class subscribes both events.** A context raises one or the other — the synchronous consumer raises `Commit`, the asynchronous one `CommitAsync` — and raising both would commit the message twice. Asserted in both directions.

**This is the caller #291 was built for.** `ITransportCommitMessage.CommitAsync` bottoms out in `IRemoveMessage.RemoveAsync`, so the ~144 lines of relational `RemoveAsync` and async delete handlers that #291 shipped as deliberately unreachable are now exercised. That was the known cost of splitting those PRs, and it is paid here.

**`IWriteMessageHistory.RecordCompleteAsync` arrives with its caller**, which is what #293 said would happen rather than adding the member early.

**Two things the tests found, not the reading.**

- A failed commit is *not* swallowed: `MessageExceptionHandler` wraps and rethrows as `MessageException`, which is what `MessageProcessingAsync` catches to roll the message back. The first version of that test asserted the opposite and failed; the contract is now pinned.
- The commit trace decorator never tagged its span with the message id, while rollback, remove and the poison handler all do — a commit span could not be tied back to its message. Fixed on both twins, the same class of gap as #289's missing `SetStatus`.

## Breaking

`IMessageContext` gains a `CommitAsync` event and `RaiseCommitAsync`; `ICommitMessage`, `ITransportCommitMessage` and `IWriteMessageHistory` gain async members. Only affects code implementing those interfaces outside this repository. The existing `Commit` event is untouched and still what the synchronous consumer raises.

## Validation

- 2,441 unit tests / 0 failed — 11 new, all failing when the commit is reverted to the blocking member
- Memory and LiteDB integration clean
- **SQLite integration 204/205.** The failure is `ConsumerHeartbeat`, which drives the *synchronous* consumer — `ProcessMessage.cs` and `MessageProcessing.cs` are untouched here — and passes 3/3 in isolation. It has now failed on two branches on two different rows, so it is filed as **#294** with both occurrences rather than re-run until green.

## Why rollback is not here

Rollback needs six transport implementations, five async command handlers with their trace decorators and DI, plus the seventh synchronous call at `ProcessMessageAsync.cs:89` that the issue's table does not list (`MessageExceptionHandler.Handle` → `IReceiveMessagesError.MessageFailedProcessing` → `RecordError`). Together with this change that is roughly 80 files, and #287 at 155 files received no LLM review at all. Splitting keeps the most intricate part of #284 reviewable.

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous message commit processing for consumers and supported transports.
  * Added awaitable commit events that run sequentially and propagate failures.
  * Added asynchronous message-history completion recording across supported storage backends.
  * Added async commit support for tracing, metrics, and in-memory queue processing.
* **Documentation**
  * Updated the changelog with asynchronous commit APIs and requirements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Fix: SQL Server and PostgreSQL left a message's error-history rows behind when it was removed with `EnableHoldTransactionUntilMessageCommitted` on (GitHub #284)
 - SQL Server, PostgreSQL and Redis async consumers no longer hold a thread recording that a message started processing, when history tracking is on (GitHub #284)
 - Async consumers no longer hold a thread committing a message. This is the per-message path, so it is the change consumers under thread-pool pressure will notice most (GitHub #284)
-- ⚠️ Custom transports and consumers must add async members: `IReceivePoisonMessage.HandleAsync`, `IRemoveMessage.RemoveAsync`, `IWriteMessageHistory.RecordProcessingStartAsync` and `RecordCompleteAsync`, `ICommitMessage.CommitAsync`, `ITransportCommitMessage.CommitAsync`, and `ITransactionWrapper.BeginTransactionAsync` for relational transports. Built-in transports are unaffected (GitHub #284)
+- ⚠️ Custom transports and consumers must add async members to `IReceivePoisonMessage`, `IRemoveMessage`, `IWriteMessageHistory`, `ICommitMessage`, `ITransportCommitMessage`, and `ITransactionWrapper`. GitHub #284 lists them; built-in transports are unaffected (GitHub #284)
 - ⚠️ `IMessageContext` gains a `CommitAsync` event and `RaiseCommitAsync`. Only affects code implementing that interface directly; the existing `Commit` event is unchanged and still used by the synchronous consumer (GitHub #284)
 - ⚠️ Custom relational transports: `IDbConnectionFactory`, `ITransactionFactory` and `ITransactionWrapper` now use `System.Data.Common` types (`DbConnection`, `DbTransaction`) rather than the `System.Data` interfaces, which have no async members. Built-in transports are unaffected (GitHub #286)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@
 - Redis async consumers no longer hold a thread while removing an expired message during a de-queue (GitHub #284)
 - Fix: SQL Server and PostgreSQL left a message's error-history rows behind when it was removed with `EnableHoldTransactionUntilMessageCommitted` on (GitHub #284)
 - SQL Server, PostgreSQL and Redis async consumers no longer hold a thread recording that a message started processing, when history tracking is on (GitHub #284)
-- ⚠️ Custom transports and consumers must add async members: `IReceivePoisonMessage.HandleAsync`, `IRemoveMessage.RemoveAsync`, `IWriteMessageHistory.RecordProcessingStartAsync`, and `ITransactionWrapper.BeginTransactionAsync` for relational transports. Built-in transports are unaffected (GitHub #284)
+- Async consumers no longer hold a thread committing a message. This is the per-message path, so it is the change consumers under thread-pool pressure will notice most (GitHub #284)
+- ⚠️ Custom transports and consumers must add async members: `IReceivePoisonMessage.HandleAsync`, `IRemoveMessage.RemoveAsync`, `IWriteMessageHistory.RecordProcessingStartAsync` and `RecordCompleteAsync`, `ICommitMessage.CommitAsync`, `ITransportCommitMessage.CommitAsync`, and `ITransactionWrapper.BeginTransactionAsync` for relational transports. Built-in transports are unaffected (GitHub #284)
+- ⚠️ `IMessageContext` gains a `CommitAsync` event and `RaiseCommitAsync`. Only affects code implementing that interface directly; the existing `Commit` event is unchanged and still used by the synchronous consumer (GitHub #284)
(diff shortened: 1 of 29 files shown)


blehnen:
Fixed in 452754a6. Both reviewers found the two real bugs, and they were mine.

## What broke

I added the `CommitAsync` subscription across the six receive classes with a script built on `re.search`, which matches **once**. Three of the six have two subscription sites:

- **Redis** has two separate methods. The subscription landed in `ReceiveMessage`, where the synchronous consumer never raises that event, and not in `ReceiveMessageAsync`, where it is the only thing that commits. Every message the asynchronous consumer processed stayed in the queue.
- **SQL Server and PostgreSQL** subscribe from one shared helper with two branches. It landed in the non-hold-transaction branch only, so the asynchronous consumer could not commit with `EnableHoldTransactionUntilMessageCommitted` on — which is why `SqlServerInboxAsyncHandlerTests.Async_Commit_BothRowsVisible` was in the failure list.

It compiled, all unit tests passed, and Memory, SQLite and LiteDB integration passed, because those three transports have a single site. The comment already sitting in `ReceiveMessageAsync` — written in #282 — says that omitting these subscriptions still compiles and silently breaks commit. That is exactly what happened.

Every path that subscribes `Commit` now subscribes `CommitAsync`, verified by counting the occurrences per file rather than trusting the build.

## One point in the review is not correct

> Do not reuse `ContextOnCommitAsync`, because it raises `CommitAsync` again.

It does not. There are two same-named members on different interfaces:

- `ICommitMessage.CommitAsync` → `context.RaiseCommitAsync()` — the one that raises.
- `ITransportCommitMessage.CommitAsync` → `IRemoveMessage.RemoveAsync(...)` — the one that removes.

`ContextOnCommitAsync` calls `_handleMessage.CommitMessage.CommitAsync(...)`, and `_handleMessage` is `ITransportHandleMessage`, whose `CommitMessage` is `ITransportCommitMessage`. So it removes; it does not re-raise. Its synchronous twin `ContextOnCommit` has the identical shape.

Reusing it in the held-transaction branch is also the *correct* behaviour rather than merely safe: `_commitConnection` is `(c, b) => _handleMessage.CommitMessage.Commit(context)`, so the synchronous held-transaction path ends in the same place, and `RemoveAsync` already deletes through `_deleteTransactionalMessageCommandAsync` and then commits `connection.Transaction` — that is #291's code. A dedicated handler would duplicate it.

## Also fixed

`CommitAsync` was never unsubscribed on cleanup in any of the six, while `Commit` always was.

## Sonar

- Two **S2077** accepted: a table name cannot be a SQL parameter, and the other interpolations in that same file are already accepted; these are the async twins of two of them.
- **S1192** fixed by naming `@PrevStatus` as a constant, which is what the file already does for its other parameters.
- **CA1861** fixed by hoisting a test's expected sequence to a `static readonly` field.

## Changelog

The custom-transport entry was 397 characters, over the 300 the guidance allows. It now names the interfaces and points at #284 for the member list, at 274.

## Not changed

**Licence headers on the two new test files.** `CLAUDE.md` scopes that rule to library and production files and says explicitly not to flag test files — amended in `0e6396e9` after the same comment on #289, #291 and #293. Every production file does comply.

## Validation

2,441 unit tests / 0 failed. Memory, SQLite and LiteDB integration clean. `NoTests.sln -c Release -p:CI=true` clean. Redis, SQL Server and PostgreSQL are Jenkins' to confirm — they are the three I cannot run locally, and the three that broke.

blehnen:
Pushed in 25567f78.

## Verified locally on the three transports that broke

SQL Server 24, PostgreSQL 30, Redis 11 asynchronous consumer tests — all passing against the real services. That is the check that should have run before the first push, not after Jenkins found the problem.

## The nitpick was already covered

Unsubscribing `_cachedCommitAsync` in cleanup landed in `452754a6`, across all six receivers.

## Coverage: the uncovered lines were worth covering, not accepting

The patch sat at 87.42% against a 90% target, and the largest gap was `CommitMessage.CommitAsync`'s catch block — the path that turns a failed commit into a `CommitException`. That is what `MessageProcessingAsync` catches to roll a message back, so it is load-bearing error handling with no test. Now covered, including that the exception carries the message id, correlation id and headers; without those the notification raised for a failed commit cannot say which message failed.

Also added:

- The commit history decorator's asynchronous member, including that a **failed history write does not turn a committed message into a failed one**. History is bookkeeping; failing to record it must not roll back work already done.
- The no-op handler's two asynchronous members.

What remains uncovered needs a real database or Redis — the relational and Redis history handlers — which the integration suites reach.

## Validation

1,150 core unit tests / 0 failed. SQL Server, PostgreSQL and Redis async consumer suites pass locally. Memory, SQLite and LiteDB integration clean. `NoTests.sln -c Release -p:CI=true` clean.

PR #296, opened 2026-09-10
feat: roll a message back without blocking on the async consumer (#284)

blehnen:
The last of the six calls #284 tracks. `MessageProcessingAsync` rolls back from three separate catch paths, all of which run in a continuation on a thread-pool thread, and each held one for a transport round trip.

With this, **#284's six rows are complete.** The seventh call I found while surveying follows separately — see the end.

## The shape of the conversion is the point

Four defects in this issue have come from writing an async copy of a synchronous method and inheriting its bugs — the missing `MetaDataErrors` delete (#291), the null-id header lookup (#291), the untagged commit span (#295), the history test that never removed its queue (#293). So nothing here is a mirrored twin. The shared part is extracted and only the execution differs:

- **The five rollback command handlers** keep one `PrepareRollback` / `PrepareStatus` pair each, used by both members. Which statement runs depends on the delayed-processing option *and* whether a heartbeat was recorded — a four-way choice two copies would be free to drift on. SQLite's fifty lines of parameter setup were lifted out of the synchronous member verbatim rather than retyped.
- **The six transport rollback implementations** share `TryBuildRollback`, which decides whether there is anything to roll back at all (`EnableDelayedProcessing || EnableHeartBeat || EnableStatus`).
- **The rollback trace decorators** share `AddTags`.

## What to look at

**One awaited handler serves every subscription site**, including the held-transaction branch where the synchronous path goes through a separate `_rollbackConnection` delegate. That is correct rather than merely convenient: `RollbackAsync` detects a held transaction itself and rolls that back instead of the message, exactly as the synchronous path ends up doing.

**Subscription counts were checked per file, before and after editing.** Two of the six receive classes have two subscription sites; getting that wrong in #295 broke async commit on three transports and was caught by Jenkins rather than locally.

**A pre-existing asymmetry, mirrored not fixed.** SQL Server wraps its held-transaction rollback in try/catch and disposes the transaction; PostgreSQL just calls `Rollback()`. That predates this work. Each transport keeps its own shape and PostgreSQL's carries a comment saying so — changing rollback behaviour on a held transaction is not something to slip into a conversion. Worth deciding separately if PostgreSQL should match.

## Two things the tests caught

- **A decorator cannot serve both handler interfaces.** Registering one type for `ICommandHandler` and `ICommandHandlerAsync` made it decorate itself on each, which SimpleInjector rejects as a cyclic graph — 26 container tests failed. There are now two classes per transport sharing one `AddTags`.
- **`MessageException` does not escape `HandleAsync`.** The catch path rethrows, but `RunAsync` swallows it deliberately ("nothing else to do, but we want to avoid the general catch below"), so the worker loop is released normally. The first version of that test asserted a throw and failed.

## Breaking

`IMessageContext` gains a `RollbackAsync` event and `RaiseRollbackAsync`; `IRollbackMessage`, `ITransportRollbackMessage` and `IWriteMessageHistory` gain async members. Only affects code implementing those interfaces outside this repository. The existing `Rollback` event is untouched and still what the synchronous consumer raises.

## Validation

- 2,456 unit tests / 0 failed — 10 new, all failing when the three await sites are reverted
- **Integration on all six transports**, asynchronous consumer and rollback suites: Memory, SQLite, LiteDB, SQL Server, PostgreSQL, Redis

## The seventh call

`ProcessMessageAsync.cs:89` calls `MessageExceptionHandler.Handle` synchronously on the async continuation, which reaches `IReceiveMessagesError.MessageFailedProcessing` and from there `RecordError` and the transport's error write. It is not in #284's table. It is a different idea from rolling a message back — moving a message to the error queue and counting retries — so it follows in its own pull request rather than being bundled here.

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous message rollback support across all supported transports.
  * Rollbacks are awaited during cancellation and error handling.
  * Added asynchronous rollback events and history recording.
  * Preserved tracing, metrics, and retry-history behavior for asynchronous rollbacks.

* **Bug Fixes**
  * Ensured rollback completion before subsequent notifications or exception handling continues.
  * Fixed rollback history recording when transport processing clears the message identifier.

* **Tests**
  * Added coverage for async event delivery, rollback completion, failure scenarios, and history recording.


blehnen:
Pushed in f4ba1bce. The review found a real bug and the coverage number pointed at why it was hidden.

## Redis never recorded rollback history

`RollbackMessageHistoryDecorator` read `context.MessageId` **after** delegating, and Redis' rollback handler clears it on its way out — *"this message should not have any more actions performed on it"*. So on that transport a rolled-back message's retry count and status were left untouched, silently.

Pre-existing on the synchronous path and inherited by the asynchronous twin, which makes it the fifth defect this conversion has copied from a synchronous original. The id is now captured before delegating, on both members, with tests that fail when it is read afterwards. Changelog entry added, since a consumer relying on rollback history on Redis has been getting nothing.

## Why nothing caught it

`RecordRollbackAsync` reached no test on any transport. History is enabled only by the SimpleHistory tests, which commit; the rollback tests run with history off. The combination this branch adds was untested — which is exactly how the defect above stayed hidden, and why the relational and Redis history handlers showed 0% on the patch.

`HistoryAsyncRollbackTest` covers it, with wrappers for the six transports. Two corrections were needed to get it right:

- A handler throwing an ordinary exception goes down the **error** path, not the rollback path. `OperationCanceledException` is what `MessageProcessingAsync` catches and rolls back — the same mechanism `HandleFakeMessagesRollback` uses.
- Requiring the message to be redelivered and completed fails on the memory transport, whose rollback is a no-op. The test asserts the history write it exists for and leaves redelivery to each transport.

To be precise about what it proves: it does not distinguish the fixed decorator on transports that keep the id, so the **unit tests** pin that behaviour and the integration test covers the path.

## Deleted LiteDb's unused rollback class

`Basic/Message/RollbackMessage.cs` is internal, registered only as its own concrete type (`container.Register<RollbackMessage>`), and nothing injects it — `LiteDbRollbackMessage` is what serves `ITransportRollbackMessage`. The container was constructing it during verification while no code path called it, which is why it read 10.71% covered right after I added async surface to it. Container verification passes without it.

## Sonar

Two fixed: the unnecessary `NpgsqlTransaction` cast (`BeginTransactionAsync` is already typed on Npgsql) and the held transaction now disposed with `DisposeAsync` in the asynchronous member.

Four accepted with reasons: **S107** is an ordinary DI constructor, the shape every command handler in this codebase has; **S2139** logs DB-timeout context then rethrows, matching the three instances already accepted here and its own synchronous twin; **S2325/CA1822** must stay instance members because they are called through the transport's `RollbackMessage` property, as their synchronous twins are.

## Not changed

**"Add parallel async interfaces or bump major."** This is one deliberate breaking release; every added member has a ⚠️ changelog entry, and #284 exists because shipping a half-async consumer is worse than either alternative. Parallel interfaces would leave the built-in transports implementing both.

## Validation

- 2,459 unit tests / 0 failed
- Integration on all six transports (asynchronous consumer + rollback): Memory 9, SQLite 21, LiteDB 19, SQL Server 34, PostgreSQL 46, Redis 13
- The new history-rollback test passes on the three transports that run without external services; the other three are Jenkins'

PR #297, opened 2026-09-10
Handle a failed message without blocking the async consumer (#284)

blehnen:
The last synchronous call on the async receive path. `ProcessMessageAsync` called `MessageExceptionHandler.Handle`, and everything under it is transport I/O: reading how many times this exception has been seen for the message, writing the count back, and moving the message to the error queue.

`IReceiveMessagesError` gains `MessageFailedProcessingAsync`; `IWriteMessageHistory` gains `RecordErrorAsync`. Both are implemented on all six transports and on the metrics, trace and history decorators. With this merged, #284's remaining row is `ISendHeartBeat`.

## What to look at

**The restructured synchronous members.** `ReceiveErrorMessage<T>` and `RedisQueueReceiveMessagesError` now share `TryGetRetryInformation`, `CanRetry` and `MovedToErrorQueue` with their async members, which meant rewriting the synchronous body around them — its `bSendErrorQueue` flag is gone in favour of returning `Retry` from the branch that sets the delay. The two are equivalent, and the synchronous `ConsumerErrorTable` tests pass on all six transports. This is deliberate rather than a mirrored twin: five defects in this conversion came from copying a synchronous method and inheriting its bugs.

**Dual-interface leaf handlers.** The relational `GetErrorRetryCountQueryHandler`, `SetErrorCountCommandHandler` and `GetErrorRecordExistsQueryHandler` implement both the sync and async handler interfaces on one class, registered for both services. A *handler* can do that safely; a *decorator* cannot — that is what produced the cyclic dependency in #296. The container tests cover it.

**`SetErrorCountCommandHandler`** needed the exists-query in async form too, which is why a third handler is in the diff.

**LiteDB** has no async API (#283), so its two error handlers are `Task.Run` wrappers over the synchronous ones, matching `MoveRecordToErrorQueueCommandHandlerAsync` from #289.

## Tests

`HistoryAsyncErrorTest`, on all six transports. Error history on the async consumer was covered nowhere: the SimpleHistory tests commit, and the error tests run with history off, so `RecordErrorAsync` reached no test on any transport — the same shape of gap that hid the Redis rollback-history defect in #296. Asserting the error status also covers the processing-start row on the relational transports, whose error update only matches a record already in Processing or Enqueued.

Async twins of the six `ReceiveMessagesErrorHistoryDecorator` tests and three for `MessageExceptionHandler.HandleAsync`. Both assert the *synchronous* member is not called, which is what proves the async path is taken rather than merely available.

Validated locally before pushing, on all six transports: `ConsumerAsyncErrorTable`, `ConsumerErrorTable` and the new `HistoryAsyncErrorTests`, plus 2,468 unit tests and the Release build.

## Summary by CodeRabbit

- **New Features**
  - Added asynchronous error handling across supported queue transports.
  - Async consumers can retry failed messages or move them to the error queue without holding a thread.
  - Asynchronous error history now records exception details, completion times, and error status.
  - Added transient-failure retry support for asynchronous queries on SQL Server, PostgreSQL, and SQLite.

- **Bug Fixes**
  - Error-queue notifications now preserve the original message ID.
  - Async error history retains the original exception details.


blehnen:
Review round pushed as `deee974d`. Three of the findings were real and are fixed; three are real but belong outside this PR and now have issues; two are the licence-header rule the repo deliberately does not apply to tests.

**Fixed**

- **Notification lost the message id** (qodo #2). Confirmed: `SetMessageAndHeaders(null, …)` nulls `MessageId`, and every built-in transport calls it once the message is in the error queue, so `Complete` built the notification from a null id. The id is captured before delegating now — the same thing `ReceiveMessagesErrorHistoryDecorator` already did, with a comment saying why. **Pre-existing on the synchronous path**, so both are fixed, with a test each asserting the notification carries the original id.
- **Async queries had no retry policy** (qodo #5). Confirmed and the best catch of the round: `RetryCommandHandlerDecoratorAsync` exists, `RetryQueryHandlerDecoratorAsync` did not, so the retry-count read was the one query on the async consumer with no transient-failure handling where its synchronous twin has one. Added for SQL Server, PostgreSQL and SQLite, sharing the existing `RetryQueryHandler` pipeline (as the async command decorator shares the command one). The dashboard's async queries pick it up too, which they should have had all along.
- **Redis metadata could be null** (CodeRabbit). `GetMetaDataQueryHandler` returns null when the hash is missing, and `metadata.ErrorTracking` would then throw out of the error handler instead of the message going to the error queue. Guarded on both paths.
- Sonar S1192: `@DurationMs` is a constant. The two S107s (nine-parameter constructors) are accepted — six injected handlers is what one class serving both paths needs, and a parameter object hides the same dependencies rather than removing them.

**Declined here, filed instead**

- **SQLite's async members run synchronously** (qodo #1) — #298. True, and true of every PR in #284: `System.Data.SQLite` inherits `OpenAsync`/`ExecuteNonQueryAsync` without overriding them. The suggested `Task.Run` wrapper does not make it asynchronous, it moves a microsecond-scale local file read onto a pool thread, which is worse. `Microsoft.Data.Sqlite` does override them and is the real option; it was measured and rejected during the performance work on other grounds.
- **Error-count check-then-write is not atomic** (CodeRabbit) — #299. Real: no unique constraint on `(QueueID, ExceptionType)` on any of the three transports. Needs an upsert *and* the constraint, so it is a schema change. This PR extracted the insert-or-update choice into a shared method; it did not change the sequence.
- **Held-transaction move commits across two connections** (qodo #3) — #300. Real, and identical in the synchronous handler — it affects the sync consumer today. Wants a forced-failure test alongside the fix.

**Not applicable**

The two licence-header comments are on test files. Per `CLAUDE.md`: *"Test projects are exempt and mostly do not carry one… Do not add headers to test files to satisfy this rule, and do not flag their absence."* `HistoryAsyncErrorTest.cs` matches its siblings in that project, none of which carry one.

On the version bump: the release happens once at the end of #284, so `Directory.Build.props` stays at `0.11.0` until then. The `ReceiveErrorMessage<T>` constructor change is resolved through DI, so it only breaks code constructing it directly — covered by the existing ⚠️ bullet about custom transports.

Validated after the changes: 2,505 unit tests; the Release build with `-p:CI=true`; and because every async query on three transports now goes through a new decorator — the full SQLite async consumer and history suites, the Dashboard API integration tests on both TFMs (219 × 2), and the async consumer, error and history tests on SQL Server and PostgreSQL.

PR #301, opened 2026-09-11
A heartbeat that neither blocks the async consumer nor holds a pool thread (#284)

blehnen:
The last row of #284. The issue's table names `ISendHeartBeat.Send` as the remaining synchronous call; reading the code, it is two problems and the send is the smaller one.

The send runs on the heartbeat scheduler's own job queue, not on the consumer's continuation — converting it alone would not have unblocked anything. What blocks the consumer is the **wait** for it: `SendHeartBeatInternal` held `_runningLocker` across the transport call, `Stop()` and `Dispose()` take that same lock, and `ProcessMessageAsync` calls `Stop()` on both failure paths and disposes the worker through `using` on **every** message. `IHeartBeatWorker` made it contractual — *"Stop blocks and does not return if the heartbeat is in the middle of updating."*

## What changed

**The wait.** `_runningLocker` becomes a `SemaphoreSlim`; `IHeartBeatWorker` gains `StopAsync` and `IAsyncDisposable`. The guarantee is unchanged — neither returns while a beat is updating — but the wait is awaited rather than blocked on. `ProcessMessageAsync` uses `await using` and `await StopAsync()`.

**The beat.** `ISendHeartBeat` gains `SendAsync`, implemented on every transport plus the metrics, policy and trace decorators, so a beat no longer occupies a pool thread for its round trip. The scheduler's job body is an `Expression<Action<…>>` and cannot hold an `await`, so the job **starts** the beat and hands its thread straight back — the semaphore bounds the beat's lifetime, and it is what `StopAsync` waits on.

## What to look at

**`HeartBeatWorker`** is the shutdown-sensitive part and deserves the closest read. The semaphore is deliberately not disposed: nothing touches `AvailableWaitHandle`, and disposing it would race a beat still on its way out.

**The SQL Server registration.** SQL Server's heartbeat statement declares `@date` itself where the relational one passes it as a parameter, and SQL Server registered only the *synchronous* handler explicitly — so the relational open-generic won the asynchronous registration and every beat failed with `the variable name '@date' has already been declared`, cancelling the worker's cancellation token. Registering SQL Server's own handler for the asynchronous interface fixes it. Same shape as the missing async retry decorator in #297: an async twin has to inherit the *registrations*, not just the code.

## Tests

`ConsumerAsyncHeartbeatTest` on all five transports that have a heartbeat. The handler blocks past two beat intervals and then reads the status a beat wrote, so a pass means a beat completed while the message was still being processed. This is what found the SQL Server defect — the existing synchronous heartbeat test asserts queue counts rather than beats, so it would have shipped silently.

`StopAsync_Waits_For_An_InFlight_Beat` holds a beat open and asserts `StopAsync` has not completed, then releases it and asserts the beat finished first — the guarantee `Stop` made, now without blocking.

Validated locally: 2,477 unit tests, the Release build with `-p:CI=true`, the new heartbeat test on SQL Server, PostgreSQL, SQLite, LiteDB and Redis, the existing synchronous heartbeat tests, and the SQLite and Memory async consumer suites.

## A correction to an earlier claim on this branch

While building this I hit `ConsumerHeartbeat` failing on SQL Server with 7 messages sent and 8 then 9 processed, and attributed it to #294 as a known flake. That was wrong, and it was this branch's own bug: the missing asynchronous registration described above killed every beat on the **synchronous** consumer too, since `HeartBeatWorker` is shared by both. A dead heartbeat means the monitor resets messages that are still being processed, and the queue hands them to a second worker. The test passes consistently now that the registration is fixed.

That failure signature is not a flake in any case. If the bad consumer had completed a message early the count would come out *low*; only a live message being reset makes it come out high. #294 has been corrected accordingly — its original SQLite sightings still need their own root cause, and the assertion there is right as written.

One thing to watch on CI: a single run of the core suite reported 12 failures that I could not reproduce in five subsequent runs, including the identical sequence. Named tests were not captured. Worth a second look if Jenkins shows anything similar.

With this merged, every row of #284 is done.

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous heartbeat sending across supported queue transports.
  * Added non-blocking heartbeat worker shutdown and asynchronous disposal.
  * Async consumers now safely wait for in-progress heartbeats before stopping or handling errors.

* **Bug Fixes**
  * Improved heartbeat handling across LiteDB, PostgreSQL, Redis, SQLite, and SQL Server.
  * Prevented tracing failures when a transport returns no heartbeat status.

* **Tests**
  * Added integration coverage for heartbeat delivery and graceful asynchronous shutdown.


blehnen:
Review round pushed as `f16fd4f3`. Both reviewers found the same real defect, and CodeRabbit found that one of my tests did not test what it claimed.

**Fixed**

- **Null heartbeat status in the trace decorator** (qodo #2, CodeRabbit). Confirmed and the most consequential of the round. `RedisQueueSendHeartBeat` returns null for a context with no message id, and `Tag` dereferenced it — so tracing turned a documented no-op into an exception, which `HeartBeatWorker` logs as a heartbeat failure and answers by cancelling the message's token. Worth noting this was **pre-existing**: the synchronous member had the identical dereference before this branch extracted the helper. Guarded in the one place, which fixes both members.
- **The fire-and-forget beat could fault unobserved** (qodo #3). `SendHeartBeatAsync` handles a failing beat, but its own failure handling can throw — a cancellation callback raising from `SetCancel` — and that fault had nobody watching. Now logged. A heartbeat that quietly stops beating is precisely what lets the monitor reset a message that is still being processed, so making these visible matters more than usual here.
- **`StopAsync_Waits_For_An_InFlight_Beat` was toothless** (CodeRabbit). Correct: NSubstitute runs a `Returns` callback inside the call, so blocking there proved only that the worker calls `SendAsync`, not that it awaits the result — which is the entire guarantee. It now returns an incomplete `TaskCompletionSource` task. I verified the new form by removing the `await` from the worker: it fails with *"StopAsync returned while a heartbeat was still updating"*, which the old form did not.
- **Changelog** split into one bullet per effect (qodo #1), plus a bullet for the tracing fix above.

**Sonar** — seven `S8949` on the semaphore waits, which failed the reliability gate. Every wait now passes `CancellationToken.None` explicitly. Passing `_cancel.Token`, the other option the rule offers, would be wrong: that token fires *when a beat has failed*, so the wait would be cancelled at exactly the moment `Stop` must not return. These waits are deliberately not cancellable, as the lock they replaced was not. Also `GC.SuppressFinalize` in the no-op's `DisposeAsync` (CA1816).

**Declined** — the licence header on the new integration test. Test projects are exempt per `CLAUDE.md`, and its siblings in that project carry none.

**One thing I checked that nobody asked about.** `lock` is reentrant on the same thread; `SemaphoreSlim` is not, so the swap could have introduced a deadlock. Every path that holds the lock: `Start` holds it across `AddUpdateJob`, but the job body returns immediately so no thread is parked; `SendHeartBeatAsync` holds it across the transport call, whose only callback is `WorkerHeartBeatNotification` — a data holder with its own private lock and no route back into the worker; the catch re-acquires only after the inner `finally` released. `Dispose` holds it across `RemoveJob`, which previously could wait on a beat running synchronously on the scheduler's thread and wanting the same lock — that deadlock is removed by the beat no longer occupying that thread.

Validated: 2,477 unit tests, the Release build, and the heartbeat tests on Redis and SQL Server (the two transports the null-status and registration fixes touch).

PR #302, opened 2026-09-11
A worker whose heartbeat has lapsed no longer keeps the message (#294)

blehnen:
The monitor resets any message whose stored heartbeat is older than the configured window and hands it to another worker. The original worker was never told — `SetCancel` fired when a beat *threw*, but not when beats simply stopped landing. So a lapse did not produce redelivery, it produced **two workers finishing the same message**. That is what `ConsumerHeartbeat` has been reporting for months as a processed count higher than the number of messages sent.

## What this changes

`HeartBeatWorker` records when a beat actually **lands** — a beat that updated no rows does not count — and checks that age on every scheduled tick, including ones where the beat is skipped. Past the expiry it trips the same cancellation token a throwing beat already used, and logs why.

⚠️ Behaviour change: a consumer that has lost its claim is now cancelled mid-message.

## What it does *not* do, and why the PR says so

The token is advisory. A handler that ignores it still runs to completion — `HandleFakeMessages` in our own integration tests does exactly that. So this makes a lapse **visible and recoverable**, not impossible. That is inherent to at-least-once delivery, and it is why preventing the lapse matters separately.

## Where the lapses actually come from

I instrumented the heartbeat on SQLite rather than reasoning about it, because two earlier hypotheses of mine were wrong:

| | |
|---|---|
| beat duration (send start → done) | p50 **39ms**, p99 1.3s, worst 1.53s |
| beats taking longer than a 3s slot | **0** |
| tick arrival interval (cron = 3.00s) | p50 2.97s, p90 3.00s, **worst 9.00s** |

All 376 ticks produced a send; none was skipped. The worker does its job in 39 milliseconds. The **scheduler** delivered one tick 9.00 seconds after the previous one, against a 10 second expiry — two slots silently missed, one second of margin left. The lapse is tick latency in the heartbeat scheduler, which dispatches beats through an internal memory queue with `ThreadsMax = 2`.

That is a design change deserving its own review, so it is **not** in this PR. Filed separately.

## Two smaller changes

- A tick arriving while a beat is still running no longer forfeits its slot; it is taken when that beat finishes, bounded to one catch-up.
- A skipped beat is logged instead of dropped silently.

Both are defensible on their own. Neither is the fix — and the measurement above is how I know: the skip path never executes in this scenario. The logging is what proved it, which is most of its value. I kept the catch-up because a slow transport can exercise it even though SQLite does not.

## Tests

Three, each verified by breaking the thing it covers:

- a claim aged past the expiry cancels processing — removing the guard fails it with *"the worker kept processing a message whose claim had lapsed"*
- a claim being kept alive is **not** cancelled — the negative case matters more, since a false positive would abort healthy work
- a tick arriving during a beat is taken afterwards — forfeiting the slot again fails it

1,191 core unit tests, all seven transport unit suites, and the sync and async heartbeat integration tests on SQL Server, PostgreSQL, SQLite, Redis and LiteDB.

## Summary by CodeRabbit

* **Bug Fixes**
  * Consumers are now cancelled when their message heartbeat becomes stale, preventing duplicate message completion.
  * Heartbeat processing recovers from scheduler ticks missed during an active update.
  * Successful heartbeat updates refresh the claim expiry window.
  * Heartbeats no longer report success after a message has been reclaimed.

* **Documentation**
  * Added an unreleased changelog entry documenting stale-heartbeat cancellation behavior.


blehnen:
Pushed as `2abeb6ee`.

**The race is real and is fixed.** The scheduler starts a tick without waiting for the previous beat, so a beat can have succeeded at the transport and not yet published its timestamp. A tick reading the claim in that window sees the old value and cancels a worker whose claim had *just been renewed* — a false positive that aborts healthy work, which is the one failure mode this feature must not have. The check now waits on the beat lock before acting, so it is ordered after any beat in flight and sees whatever that beat published. `ClaimHasLapsed` became a pure read and the cancelling moved into the confirm step.

**On the coverage number** — the project figure did not drop; it is 90.55%, against 90.63% on master. The 80% was *patch* coverage, and the ten uncovered lines were the branches this change exists for: the ones that report a lost claim. That is worth fixing on its own merits rather than for the number, so:

- Redis reporting no heartbeat when the message is no longer in the working set, on both members, plus the happy path and the no-id path. That needed a `GetDb()` seam — `IRedisConnection` exposes a concrete multiplexer, which is the same reason `WriteMessageHistoryHandler` already has one.
- `RedisQueueSendHeartBeat` turning a zero timestamp into "no heartbeat" rather than one dated at the epoch.
- The tick-during-a-landing-beat case above.

**One correction worth stating.** My first attempt at verifying that last test was invalid and I nearly reported it as proof. Stubbing out `ClaimHasLapsed()` cancels nothing, because the refactor had already moved the cancel out of it — so the test passed against a broken control. Checked properly, against what the code did *before* the fix (cancel on the first reading), it fails with *"a worker was cancelled on a stale reading taken while a heartbeat was landing"*.

1,192 core tests, all seven transport unit suites, the Release build, and the heartbeat integration tests on Redis, SQLite and SQL Server.

Issue #298, opened 2026-09-10
SQLite: the async data path is synchronous, because System.Data.SQLite does not override the async members


blehnen:
`System.Data.SQLite` inherits `DbConnection.OpenAsync`, `DbCommand.ExecuteNonQueryAsync`, `ExecuteReaderAsync` and `DbDataReader.ReadAsync` without overriding them, so the base-class implementations run the synchronous work on the calling thread and hand back a completed task. Every `await` on the SQLite async path therefore completes synchronously.

This affects all of #284's work, not one PR of it: the delete handlers (#291), the history writes (#293), the commit path (#295), rollback (#296) and the error path (#297) all reach SQLite through those members.

The practical impact is small — SQLite is a local file, so the blocking call is microseconds rather than a network round trip, and unlike the Redis case there is no sync-over-async completion to starve the pool. That is why it was left alone rather than wrapped.

Options, if it turns out to matter:

- `Task.Run` wrappers over the synchronous handlers, the way LiteDB's handlers do it (#283). This does not make the work asynchronous; it moves the block onto a pool thread, which is worse for a microsecond-scale file read.
- `Microsoft.Data.Sqlite`, which does override them. Measured during the SQLite performance work and rejected on throughput grounds (15.5us against our 14.8us) — but that measurement was about the synchronous path, and the async story would be a separate reason to revisit it.

Raised by review on #297. No action taken there; documenting the constraint so it is not re-derived.

blehnen:
Checked the second option before documenting it, and it does not hold. Reflecting over the shipped assemblies, with Npgsql as a control:

- **System.Data.SQLite 1.0.119** — `SQLiteConnection`, `SQLiteCommand`, `SQLiteDataReader` and `SQLiteTransaction` declare none of `OpenAsync`, `ExecuteNonQueryAsync`, `ExecuteReaderAsync`, `ReadAsync`, `CommitAsync`, `RollbackAsync`. Confirms the first half of this issue.
- **Microsoft.Data.Sqlite 9.0.0** — declares only `ExecuteReaderAsync` and `ExecuteDbDataReaderAsync`.
- **Npgsql 8.0.8** (control) — declares the whole set.

[...]

Issue #292, opened 2026-09-09
Relational delete handlers report Removed for a message that was not there


blehnen:
`DeleteMessageCommandHandler` and `DeleteTransactionalMessageCommandHandler` both return a hardcoded `1` rather than the affected-row count:

```csharp
_prepareCommand.Handle(command, commandSql, CommandStringTypes.DeleteFromQueue);
commandSql.ExecuteNonQuery();   // count discarded
...
trans.Commit();
return 1;
```

`RemoveMessage.Remove` turns anything above zero into `RemoveMessageStatus.Removed`, so a message id that has no queue row — already removed, or never existed — is reported as `Removed` instead of `NotFound`.

## Who sees it

Anything that acts on the status rather than ignoring it. `ClearExpiredMessages` and `ClearErrorMessages` count what they removed, and the dashboard delete surfaces the result to a user.

## Scope

Long-standing on both handlers and both transports, and the same shape in the async twins added in #291, which mirror them deliberately. Fixing it means deciding which statement's count is authoritative — the queue row is the natural answer, since the metadata and status rows are derived — and then whether `Remove` reporting `NotFound` where it used to report `Removed` needs a changelog ⚠️ for callers that branch on it.

Not folded into #291: that pull request converts the call to be awaitable and mirrors existing behaviour on purpose, and changing what `Remove` returns is a behaviour change for the synchronous path that deserves its own review.

Found by review on #291.

blehnen:
Fixed by #305.

Issue #288, opened 2026-09-09
SQLite async send handlers block on synchronous ADO calls (6 x Sonar S6966)


blehnen:
## What

SonarCloud reports six S6966 findings on the SQLite async send handlers. All six are `async` methods that block on the synchronous ADO call where an awaitable one now exists:

| File | Line | Finding |
|---|---|---|
| `SendMessageCommandBatchHandlerAsync.cs` | 111 | Await `OpenAsync` instead |
| `SendMessageCommandBatchHandlerAsync.cs` | 122 | Await `CommitAsync` instead |
| `SendMessageCommandBatchHandlerAsync.cs` | 128 | Await `RollbackAsync` instead |
| `SendMessageCommandHandlerAsync.cs` | 138 | Await `OpenAsync` instead |
| `SendMessageCommandHandlerAsync.cs` | 210 | Await `CommitAsync` instead |
| `SendMessageCommandHandlerAsync.cs` | 226 | Await `DisposeAsync` instead |

## Why they only appeared now

These calls are not new, and neither is the blocking. Sonar could not report them before #287 because the connection and transaction were typed as `IDbConnection` and `IDbTransaction`, which have no asynchronous members — there was no `OpenAsync` for the rule to point at. Moving the abstraction to `System.Data.Common` made six pre-existing sync-over-async calls visible in code that was already `async`.

That is a useful independent signal that #286 did what it was for.

## Why they were left out of #287

#287 is an enabling refactor: 155 files, no behaviour change, and — because CodeRabbit's free plan declines pull requests over 100 files — no LLM review. Adding six behaviour changes to it would have been the wrong trade.

They also need validation this repository cannot do casually. `SendMessageCommandBatchHandlerAsync` is the batch send path, where transaction atomicity has produced real shipped bugs before (see #250 and the `XACT_ABORT` lesson), and confirming these changes needs SQL Server and PostgreSQL integration runs.

## What to do

Convert the six, then validate on Jenkins rather than locally. Points to watch:

- **`CommitAsync` and `RollbackAsync` are not universally available.** They exist on `DbTransaction` from .NET Standard 2.1 onward, but a provider may not override them, in which case the base implementation runs the synchronous version on the calling thread and nothing is gained. Confirm per provider rather than assuming; the same caution that applies to `DbConnection.OpenAsync`.
[...]

PR #305, opened 2026-09-11
Three remaining correctness and blocking issues (#288, #290, #292)

blehnen:
Three small, independent fixes from the release list, batched to save build cycles. Each is self-contained; the diff separates cleanly by file.

## #292 — a delete reported success for a message that was not there

All four relational delete handlers returned a hardcoded `1`, and `RemoveMessage.Remove` turns anything above zero into `Removed`. So an id with no queue row — already removed, or never present — came back as `Removed`. That is visible: `ClearExpiredMessages` and `ClearErrorMessages` count what they removed, and the dashboard delete shows the result to a user.

They now return the affected-row count from the **queue** delete. That is the authoritative statement — meta data, status and error rows are derived from it — and it matters that the status delete runs *after*, so it must not become the answer. Covered by a test for that case specifically.

⚠️ Behaviour change for callers that branch on the status, and a changelog entry to match.

## #290 — NOSCRIPT recovery blocked the async path

`BaseLua.TryExecuteAsync` awaits the evaluation but recovered by calling the **synchronous** `LoadScript()`, which loads onto every endpoint in turn. A Redis restart, failover or `SCRIPT FLUSH` parked a thread-pool thread for the whole sweep — on the path that exists to avoid exactly that, and it affects every caller of `TryExecuteAsync`, including the async send, de-queue and poison-message paths.

Adds `LoadScriptAsync`, sharing the prepare and endpoint lookup with the synchronous loader, which stays for `TryExecute`.

## #288 — SQLite async sends blocked on synchronous ADO

The six Sonar S6966 findings, plus the two `BeginTransaction` calls of the same kind that the issue did not enumerate.

**What this buys today, stated honestly: nothing measurable.** System.Data.SQLite does not override the inherited async members, so these still complete synchronously — that is #298. The value is that the code is correct in shape, and that the findings become real signal rather than noise. It is also why this one is grouped with the others rather than sold as a performance change.

## Tests

`DeleteMessageCommandHandlerTests` — three cases, including with a status table. Verified by restoring the hardcoded `1`: the missing-message test then fails with *"a message that was not there was reported as removed"*.

The Lua and SQLite send changes are covered by existing integration tests rather than new ones, since both are shape changes on paths with heavy coverage already.

Validated: 1,192 core unit tests and all seven transport unit suites; the Release build with `-p:CI=true`; and integration — SQLite consumer, async consumer and producer suites (164), Redis async consumer (12, the Lua paths), SQL Server async consumer (25, the delete counts).

## Summary by CodeRabbit

- **Bug Fixes**
  - Removing a message that no longer exists now correctly reports `NotFound` across SQL Server, PostgreSQL, and SQLite.
  - Expired-message cleanup, error-message cleanup, and dashboard deletion counts no longer report phantom removals.
  - Redis asynchronous operations recover from `NOSCRIPT` without blocking threads.
  - SQLite asynchronous message sending now performs connection and transaction operations asynchronously, improving responsiveness.


blehnen:
Pushed as `eccb4d3a`.

**The changelog comment is right and the bullet is gone.** It claimed SQLite async sends no longer block while opening or committing. They still do — System.Data.SQLite does not override the inherited asynchronous members, which is #298, and which this PR's own description already said. Writing a bullet that contradicts the description was the mistake; the changelog's only test is *does a consumer act differently knowing this*, and the answer is no until #298. The commit carries the reasoning instead.

**On patch coverage** — project is 90.56%, so the 71.7% is the patch. Two files:

- `BaseLua` at 52%, which is nearly all of it. The uncovered code is the NOSCRIPT recovery path itself. Forcing a NOSCRIPT means a `SCRIPT FLUSH` against a live server, and that needs an admin connection — the Redis integration tests connect without `allowAdmin`, so nothing can currently reach it. Worth saying plainly: **the synchronous recovery path has never been covered either**, so this is a pre-existing gap that my change made visible rather than one it introduced. I have covered what is reachable — both loaders being called with no connection, which is the state they are in while one is coming back after a failover — and the test says why the rest is missing rather than leaving it implied.
- `SendMessageCommandBatchHandlerAsync`, one line: the rollback in the failure branch.

If you want the recovery path genuinely covered, the way to do it is an integration test on a connection opened with `allowAdmin=true` that loads a script, flushes, and re-executes through `TryExecuteAsync`. That is a change to the Redis test connection setup, so I did not fold it into this PR — say the word and it can be its own piece of work, and it would cover the synchronous path at the same time.

1,192 core tests, 211 Redis and 265 relational unit tests green.

Issue #300, opened 2026-09-10
Held-transaction move to the error queue commits the error row on a second connection


blehnen:
With `EnableHoldTransactionUntilMessageCommitted` on, `MoveRecordToErrorQueueCommandHandler` (and its async twin) opens its own connection to insert the error row and commits it there, then deletes the source record inside the held receive transaction. If the delete or the outer commit fails, the error copy is already committed and the source message rolls back into the queue — so the message exists in both places and can be processed again.

Both the synchronous and asynchronous handlers have this shape, so it affects the sync consumer today; #297 only made the async twin reachable, and review flagged it there.

Fixing it means giving the move one recoverable outcome — using the held transaction for the whole move if the cross-connection read is not actually required, or adding compensating cleanup if it is. Either way it wants a forced-failure integration test (dropping the MetaData table mid-move is the trick used elsewhere in this repo).

blehnen:
## The cross-connection read is required, and here is the proof

I tried the obvious fix — run the error insert on the held connection and transaction, so the whole move commits or rolls back together — and it breaks PostgreSQL. Worth recording in detail, because the current shape looks like an oversight and is not.

**SQL Server:** 9/9 error tests pass with the insert moved onto the held transaction.

**PostgreSQL:** 5 of 9 fail, with zero rows reaching the error table. The original code passes the same tests 5/5, so the change is the cause.

The reason is in the receive statement. With `EnableHoldTransactionUntilMessageCommitted`, PostgreSQL dequeues with:

```sql
delete from MetaData q using Queue qm where q.QueueID in ( ... )
```

The metadata row is **deleted as part of the receive**, inside the held transaction, and stays uncommitted until the message is committed. So:

- Inside that transaction, `INSERT INTO MetaDataErrors ... SELECT ... FROM MetaData WHERE QueueID = @id` sees the row already deleted — zero rows, the handler's `if (iCount != 1) return;` fires, and the message is never moved to the error queue at all.
- On a **separate** connection, the uncommitted delete is not visible, the row is still readable, and the copy succeeds.

So the second connection is deliberately reading the pre-delete snapshot. It is load-bearing, not an accident, and sharing the transaction cannot work while the error copy is built by selecting from a row the same transaction has already deleted.

## What that leaves

[...]

PR #306, opened 2026-09-11
Make the move to the error queue atomic under a held transaction (#300)

blehnen:
With `EnableHoldTransactionUntilMessageCommitted`, the error copy was inserted on a connection of its own, with no transaction, so it committed on the spot. The source delete and the commit then ran on the **held** transaction. If either failed, the error copy stayed while the message rolled back into the queue — the same message in both places, free to be processed again.

## The obvious fix does not work, and that is the interesting part

Running the insert on the held transaction gives **SQL Server 9/9 passing and PostgreSQL 5 of 9 failing, with nothing reaching the error table.** The two transports claim a message differently:

| | how a message is claimed under a held transaction |
|---|---|
| SQL Server | `update cte set queueid = QueueID` — a no-op update purely to take the row lock. The metadata row survives. |
| PostgreSQL | `delete from MetaData q using Queue qm …` — the row is **deleted** as part of the receive. |

The error copy is built by selecting from that row. Inside the transaction that deleted it there is nothing to copy, `if (iCount != 1) return;` fires, and the message never reaches the error queue at all. **The second connection was reading the pre-delete snapshot deliberately** — it looks like an oversight and is not.

## What this does instead

Makes the two transports claim a message the same way. PostgreSQL's receive now takes the lock with a no-op update rather than deleting, exactly as SQL Server does, and the commit path deletes the row for real — which it already did on both, through `DeleteTransactionalMessageCommandHandler`.

The claim is unchanged in substance: another consumer's `FOR UPDATE SKIP LOCKED` skips a locked row as readily as a deleted one, and a rollback releases the lock where it used to undo the delete. An uncommitted delete was never visible to other connections either, so nothing outside the transaction sees a difference.

**Only the held-transaction branch changes.** Without a status column and without a held transaction, deleting during the receive *is* the consumption — that path is untouched.

With the row surviving, the insert moves onto the held transaction and the move finally has one outcome.

## What to look at

This changes the **receive statement** on PostgreSQL, which is the hottest path in the library, so it got a wide sweep rather than a targeted one:

- PostgreSQL: **9/9** error tests (5/9 with the naive fix), **64** consumer, **111** rollback, producer, route and job tests
- SQL Server: 9/9 error tests, confirming the shared handler still behaves
- 1,192 core unit tests, all relational unit suites, Release build with `-p:CI=true`

Unit tests pin the invariant directly: the insert runs on the held transaction, no second connection is opened, and nothing is committed when the message was not there to copy.

SQLite is unaffected — it hard-wires `EnableHoldTransactionUntilMessageCommitted` to `false`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction handling when moving failed messages to the error queue.
  * Prevented duplicate message processing by ensuring queue removal and error-queue transfer succeed or fail together.
  * Improved PostgreSQL message locking when held transactions are enabled.
* **Tests**
  * Added coverage for transaction reuse, connection handling, commit behavior, and missing-message scenarios.
* **Documentation**
  * Updated the changelog with the transaction consistency fix.


blehnen:
Both addressed.

**The changelog entry was describing the fix rather than the effect** — "now part of the same transaction as removing it from the queue" is how it works, not what a consumer notices. Rewritten to lead with the symptom: a message that could end up in the error queue *and* back in the queue, and be processed a second time.

**The retry finding is real, and I have filed it as #309 rather than folding it in — because it predates this change and is wider than this handler.**

Precisely: before this PR the error-queue move ran its insert on a connection of its own and the delete and commit on the held transaction. After a retryable failure PostgreSQL aborts that transaction, so a retry would re-run an insert that had **already committed** — producing a duplicate error row or a key violation — and then hit the aborted transaction again and fail anyway. Retries on this path have never been able to complete the move. What this PR changes is that the insert no longer commits independently, so the futile attempts stop leaving duplicate error rows behind. The delay that remains was already there.

It is also not specific to this handler. `DeleteTransactionalMessageCommandHandler` and its async twin run entirely inside the caller's transaction and carry the same retry decorator, and have done all along.

Fixing it properly is more than a flag, which is the other reason it is its own issue. `IRetrySkippable` already expresses exactly this idea with a per-instance `SkipRetry`, but the **no-output** retry decorators do not honour it — only the with-output one does, on the grounds that the check would be dead code — so it would need adding across three transports, synchronous and asynchronous. And the command object the decorator inspects cannot tell whether a transaction is held: `MoveRecordToErrorQueueCommand` is built by `ReceiveErrorMessage`, which is transport-neutral and knows nothing about `EnableHoldTransactionUntilMessageCommitted`. That knowledge lives in the handler and the options. #309 lays out the three ways round it.

Impact meanwhile is bounded: the outcome is still correct, just delayed by a retry pipeline that cannot succeed, and only for held-transaction queues on SQL Server or PostgreSQL.

CI was green on the previous commit — Jenkins, Sonar, CodeRabbit and both codecov checks.

Issue #299, opened 2026-09-10
Relational error-count write is check-then-write, with no unique constraint behind it


blehnen:
`SetErrorCountCommandHandler` asks `GetErrorRecordExistsQuery` whether a row exists for `(QueueID, ExceptionType)` on one connection, then runs either `InsertErrorCount` or `UpdateErrorCount` on another. The `ErrorTracking` table has no unique constraint on that pair on SQL Server, PostgreSQL or SQLite, so two failures of the same message with the same exception type, arriving close enough together, can both see "no row" and both insert. Later updates then increment one of the two, and the retry count the error path reads back is lower than the real number of attempts — the message gets more retries than configured.

The window is narrow: it needs the same message failing twice concurrently, which means two workers holding it at once. That is possible with a heartbeat reset while the first worker is still running.

The fix is an atomic upsert per transport plus the unique constraint to make it enforceable, which is a schema change — hence a separate issue rather than a fix inside #297. A concurrent-attempt test belongs with it.

Pre-existing, and unchanged by #297: that PR extracted the insert-or-update choice into a shared method so the sync and async members could not diverge, but did not alter the sequence. Raised by review there.

blehnen:
Fixed by #307 (merged as 6f6014c3).

New queues carry a unique index on `ErrorTracking (QueueID, ExceptionType)` and count with a single statement — `ON CONFLICT` on PostgreSQL and SQLite, `UPDATE … WITH (UPDLOCK, HOLDLOCK)` plus a conditional insert inside one transaction on SQL Server.

Queues created before the index exists keep the old check-then-write, since the library does not upgrade schemas. That fallback is a stopgap and #308 records exactly what to delete once upgrades exist.

PR #307, opened 2026-09-11
Count message errors in one statement, where the schema allows it (#299)

blehnen:
`SetErrorCountCommandHandler` asked whether a row existed for `(QueueID, ExceptionType)` and then inserted or updated, with nothing enforcing one row per pair. Two failures of the same message arriving together could each see no row and each insert one. The retry count then reads lower than reality — so a message gets more attempts than configured, and a poison message can loop instead of reaching the error queue.

## What changed

New queues get a unique index on `(QueueID, ExceptionType)`, and the count becomes a single statement:

| | |
|---|---|
| PostgreSQL, SQLite | `INSERT … ON CONFLICT (QueueID, ExceptionType) DO UPDATE` |
| SQL Server | `UPDATE … WITH (UPDLOCK, HOLDLOCK)` then insert when no row was touched, inside one explicit transaction |

## Existing queues

The library does not upgrade schemas — a user drains and re-creates — so an older error tracking table has no such index, and `ON CONFLICT` against a table without one throws rather than degrading. The queue is therefore asked once whether the index is there and keeps the old check-then-write when it is not. The answer is cached: the schema does not change underneath a running consumer, and this sits on the failure path. A look-up that *fails* is not cached, so one transient outage cannot pin a queue to the old path for its lifetime.

**Adding the index cannot break an existing deployment.** `CreateQueue` returns `AlreadyExists` without touching anything when the tables are present, so the index is only ever created as part of a new, empty table. An upgrade that added it to a live table would have to de-duplicate first — a queue that ran on an older version may hold exactly the duplicate rows this bug produces. That is written into the upgrade issue (#308).

## Where to look

The detection query is the part worth review. It matches the index by **shape** rather than by name, per transport:

- unique, and not partial or filtered — a predicate means the guarantee does not hold for the rows outside it, and `ON CONFLICT` cannot infer such an index anyway
- exactly two **key** columns, `QueueID` and `ExceptionType`, compared case-insensitively — an `INCLUDE`d column is stored by the index but carries no uniqueness
- the table resolved the way the transport's own statements resolve it: `to_regclass` on PostgreSQL (schema-qualified queue names are legal, and identifiers are folded and truncated), `OBJECT_ID` on SQL Server, `pragma_index_list` on SQLite

Matching by name does not work here: PostgreSQL lower-cases and truncates, SQLite appends the table name to whatever name it is given, and SQL Server scopes names per table.

## Tests

Unit: both branches, that the schema is asked about once, and that a failed look-up falls back for that call only and is retried on the next.

Against real databases, on all three transports:

- a new queue reports the index and counts through the single statement; drop the index and the fallback still counts
- 32 workers failing the same message at once produce one row with a count of 32
- a unique index on `QueueID` that merely `INCLUDE`s `ExceptionType` is not accepted as the key (PostgreSQL, SQL Server — SQLite has no included columns)

Each asserts what the queue *decided*, not only the resulting count. Both paths count correctly when nothing is racing, so a count-only assertion passes either way.

Validated locally: relational unit suites, the integration tests above on SQL Server, PostgreSQL and SQLite, and a `Release -p:CI=true` build. The full error-path integration sweep runs on Jenkins.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ﻿### Unreleased
+- Fix: counting how many times a message has failed is now one statement on SQL Server, PostgreSQL and SQLite. Two failures arriving together could each add a row, and the retry count then read low - so a message got more attempts than configured, and a poison message could loop instead of reaching the error queue (GitHub #299)
+- New queues get a unique index on the error tracking table. Existing queues keep working unchanged: the queue is asked once whether the index is there and falls back to the previous behaviour when it is not. Re-create a queue to get the guarantee (GitHub #299)
 - Redis consumers no longer hold a thread through the de-queue. Under thread-pool pressure, where a consumer previously timed out waiting on Redis, it now keeps working (GitHub #256)
 - ⚠️ Custom transports and consumers must add async members: `IReceiveMessages.ReceiveMessageAsync`, `IQueueWait.WaitAsync`, and `IRedisQueueWorkSub.WaitAsync` for Redis. Built-in transports are unaffected (GitHub #256)
 - ⚠️ `IMessageProcessing.Handle()` is now `HandleAsync()` and returns a task. Only affects code implementing that interface directly (GitHub #256)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 ﻿### Unreleased
-- Fix: counting how many times a message has failed is now one statement on SQL Server, PostgreSQL and SQLite. Two failures arriving together could each add a row, and the retry count then read low - so a message got more attempts than configured, and a poison message could loop instead of reaching the error queue (GitHub #299)
-- New queues get a unique index on the error tracking table. Existing queues keep working unchanged: the queue is asked once whether the index is there and falls back to the previous behaviour when it is not. Re-create a queue to get the guarantee (GitHub #299)
+- ⚠️ Fix: two failures arriving at once each added an error row, so a message got more retries than configured and could loop instead of reaching the error queue. Re-create an existing SQL Server, PostgreSQL or SQLite queue to pick up the fix (GitHub #299)
 - Redis consumers no longer hold a thread through the de-queue. Under thread-pool pressure, where a consumer previously timed out waiting on Redis, it now keeps working (GitHub #256)
 - ⚠️ Custom transports and consumers must add async members: `IReceiveMessages.ReceiveMessageAsync`, `IQueueWait.WaitAsync`, and `IRedisQueueWorkSub.WaitAsync` for Redis. Built-in transports are unaffected (GitHub #256)
 - ⚠️ `IMessageProcessing.Handle()` is now `HandleAsync()` and returns a task. Only affects code implementing that interface directly (GitHub #256)
--- a/Source/DotNetWorkQueue.Transport.PostgreSQL.Integration.Tests/Basic/SetErrorCountAgainstBothSchemasTests.cs
+++ b/Source/DotNetWorkQueue.Transport.PostgreSQL.Integration.Tests/Basic/SetErrorCountAgainstBothSchemasTests.cs
@@ -4,10 +4,10 @@ using DotNetWorkQueue.Transport.RelationalDatabase.Basic.Query;
 using DotNetWorkQueue.Transport.Shared;
 using DotNetWorkQueue.Transport.Shared.Basic.Command;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using DotNetWorkQueue.Transport.SQLite.Basic;
-using System.Data.SQLite;
+using DotNetWorkQueue.Transport.PostgreSQL.Basic;
+using Npgsql;
 
-namespace DotNetWorkQueue.Transport.SQLite.Integration.Tests.Basic
+namespace DotNetWorkQueue.Transport.PostgreSQL.Integration.Tests.Basic
 {
     /// <summary>
     /// Counting errors against a real database, with and without the unique index.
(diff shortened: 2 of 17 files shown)


blehnen:
Review round pushed (rebased onto master).

**The one that mattered: detection was broken on PostgreSQL.** The check compared the index name the schema asked for against the catalog. PostgreSQL folds unquoted identifiers to lower case and truncates at 63 bytes, so it never matched — detection always answered no, and this change did nothing on that transport. Nothing failed, because the fallback still counts correctly when nothing is racing.

Detection now asks about the index's **shape** on all three transports — unique, exactly two columns, `QueueID` and `ExceptionType` — so a folded, truncated or differently-suffixed name cannot hide it, and a same-named non-unique index cannot be mistaken for it.

**SQL Server upsert was not atomic.** Two statements in autocommit release the range lock between them. Now wrapped in an explicit `begin transaction … commit transaction`, which is what makes the `UPDLOCK, HOLDLOCK` + conditional insert safe.

Also fixed:
- enum members **appended**, not inserted — `CommandStringTypes` is public with implicit values, so inserting renumbers every member after it (binary compatibility)
- probe now runs **before** the outer connection is opened, on both the sync and async paths — it could otherwise deadlock a pool of one
- `Lazy<bool>` is `PublicationOnly` with a `catch → false`, so a transient outage during the probe, or a custom relational transport whose command cache has no entry for the new query, falls back instead of throwing
- the query carries only the table name; the prepare handler binds only `@Table`
- SQL Server index name dropped the table suffix — index names are table-scoped there, and the suffix pushed the identifier past 128 characters for queue names the validator allows

**On #7 (only SQLite had a real-database test).** Fixed, and the test was wrong in a second way: it only asserted the resulting count, which passes on *both* branches. `SetErrorCountAgainstBothSchemasTests` now exists for PostgreSQL and SQL Server as well, and each asserts **detection** either side of dropping the index. Teeth-checked — restoring the name comparison on PostgreSQL fails the new assertion, which is exactly the bug above that a count-only test slept through.

**Changelog** (#8/#9/#10): collapsed to one bullet on the observable behaviour plus what an existing queue owner has to do.

Verified locally against live databases: error-path integration (error table / poison / rollback) on SQLite 19/19, PostgreSQL 41/41, SQL Server 31/31; the new both-schemas test passes on all three; relational unit suites 269/198/183/242; `Release -p:CI=true` clean.

blehnen:
Second round — four fixed, one declined.

**Schema-qualified table names (PostgreSQL, major).** Correct, and it is worse than stated: a dot is legal in a queue name, so `app.Queue` gives the table `app.QueueErrorTracking`, and the probe both forced `nspname = 'public'` and compared the *whole* qualified name against `relname`. Rather than parse the name, the probe now resolves it with `to_regclass(@Table)` — the same folding, the same 63-byte truncation, a schema-qualified name honoured instead of assumed away, and `null` rather than an error when the table is not there. SQL Server already had this right via `OBJECT_ID(@Table)`.

**Partial unique indexes (major).** Correct. Added `ix.indpred IS NULL` (PostgreSQL) and `il."partial" = 0` (SQLite), and `i.has_filter = 0` on SQL Server for the same reason — a filtered index fits the shape but guarantees nothing about the rows outside its filter, which is the guarantee being asked about.

**Caching a failed probe as `false` (major).** Correct, and the better framing is yours: one transient blip would pin the queue to the racy path for the rest of its life. The look-up now throws, and each use catches it, so the next failed message asks again — `PublicationOnly` does not remember an exception. Two unit tests, one of which asserts the second failure re-attempts the look-up; teeth-checked against the old catch-inside-the-factory (it reports 1 attempt, not 2).

**Concurrent first-error coverage.** Added on all three transports: 32 workers failing the same message at once, asserting exactly one row with `RetryCount = 32`.

32 rather than a handful because the first attempt at 8 was worth reporting: with the fix removed, PostgreSQL and SQL Server both failed, and **SQLite passed** — it serialises writers, so the interleaving never happened and the test proved nothing there. At 32 all three fail without the fix. With the fix, 4 consecutive runs green on each.

**Declined — licence headers on the three new test files.** Test projects are exempt; see `CLAUDE.md` → Conventions: *"Test projects are exempt and mostly do not carry one; #206 scoped its header sweep to production files deliberately. Do not add headers to test files to satisfy this rule, and do not flag their absence."*

Verified: relational unit suites 275/198/183/242, error-path integration (error table / poison / rollback) SQLite 19/19, PostgreSQL 41/41, SQL Server 31/31, both new tests green on all three transports, `Release -p:CI=true` clean.

blehnen:
Third round — one finding, valid on both transports.

**Included columns counted as key columns.** Correct. A unique index on `QueueID` alone that merely `INCLUDE`s `ExceptionType` has both names and is unique, so the probe said yes to a question it does not answer: it guarantees one row *per message*, not one per message and exception type. The single statement would then fail on the second exception type for a message — PostgreSQL cannot infer a conflict target it has no unique key for, and SQL Server’s insert collides with the constraint on `QueueID`.

- SQL Server: `ic.key_ordinal > 0` on the count and both column checks (`sys.index_columns` lists INCLUDE columns with `key_ordinal` 0)
- PostgreSQL: `ix.indnkeyatts = 2` instead of `array_length(ix.indkey, 1)`, and the key columns read as `ix.indkey[0]`/`ix.indkey[1]` — key columns come first in `indkey`, included ones after. `indnkeyatts` is PostgreSQL 11+, comfortably below what Npgsql 10 supports.
- SQLite: no change, it has no included columns.

`DoesNotMistakeAnIncludedColumnForAKeyColumn` on both transports drops the real index, creates the misleading one and asserts the probe says no. Teeth-checked — reverting either change fails it.

Verified locally: relational unit suites 275/198/183/242, the both-schemas suites green on all three transports, `Release -p:CI=true` clean. Leaving the full error-path integration sweep to Jenkins on this push.

Issue #304, opened 2026-09-11
Timestamps are taken from DateTime.UtcNow in places that bypass the configured time provider


blehnen:
Found while fixing #294, where `HeartBeatWorker` was judging a claim's freshness with `DateTime.UtcNow` while the monitor ages the stored heartbeat using `IGetTime`. That one is fixed in #302; this issue is the rest of them.

## The point

The queue lets a user choose where time comes from. `IGetTime` is registered per transport:

| transport | provider |
|---|---|
| SQL Server | `SqlServerTime` |
| PostgreSQL | `PostgreSqlTime` |
| Redis | `GetRedisTimeFactory` → Redis server time (or SNTP, or local) |
| core default | `LocalMachineTime` |

So on the relational transports the default is already the **database server's** clock, and a user can opt into SNTP. Code that calls `DateTime.UtcNow` directly ignores all of that. It is not that local time is wrong — the default provider is local time on some transports — it is that the choice is silently not honoured, so a queue configured for NTP gets NTP in some places and the machine clock in others.

It is also cheap to do properly: `BaseTime.GetCurrentUtcDate()` adds a cached offset to the local clock and only refreshes periodically, so there is no per-call round trip.

## Where it happens

Production code only; the time providers themselves are excluded, since `DateTime.UtcNow` is exactly what they are supposed to use.

| file | uses | note |
|---|---|---|
| `Transport.RelationalDatabase/Basic/WriteMessageHistoryHandler.cs` | 9 | `EnqueuedUtc`, `StartedUtc`, `CompletedUtc` and the durations derived from them. No time source injected. Worth doing first: on SQL Server and PostgreSQL the rest of the queue is on server time, so history written from an app machine with a different clock can report durations that are wrong, or negative |
| `Transport.Redis/Basic/WriteMessageHistoryHandler.cs` | 9 | same, and Redis defaults to Redis server time |
| `Transport.LiteDB/Basic/WriteMessageHistoryHandler.cs` | 6 | |
| `Transport/Memory/Basic/WriteMessageHistoryHandler.cs` | 6 | |
| `Transport.LiteDB/Basic/CommandHandler/SendMessageCommandHandler.cs` and `…Async`, `…BatchShared` | 9 | the send path |
[...]

PR #310, opened 2026-09-11
History timestamps come from the configured time provider (#304)

blehnen:
The queue lets a user choose where time comes from — `IGetTime` is registered per transport, and on SQL Server and PostgreSQL the default is the **database server's** clock, not the local one. The four `WriteMessageHistoryHandler`s ignored that and called `DateTime.UtcNow`, 36 calls between them.

The effect is not that local time is wrong — it is the default on some transports — but that these timestamps land beside data written on a different clock. History written from an application machine that drifts from the database server can record a `StartedUtc` later than its `CompletedUtc`, and `DurationMs` is derived by subtracting the two.

## What changed

`Transport.RelationalDatabase`, `Transport.Redis`, `Transport.LiteDB` and `Transport/Memory` history handlers take `IGetTimeFactory` and hold the `IGetTime` it creates, the same shape `HeartBeatWorker` uses. `BaseTime` caches an offset against the local clock and refreshes periodically, so none of these calls costs a round trip.

## Where to look

The constructor change is the only behavioural decision here: the `IGetTime` is resolved once and held, rather than resolved per call. These handlers are registered `LifeStyles.Singleton`, and container verification passes on every transport.

## Tests

The existing unit tests constructed these handlers directly and asserted only that a timestamp was present and not in the future — which `DateTime.UtcNow` satisfies, so they could not tell the two sources apart. They now drive a clock fixed at a date nowhere near the machine's and assert the stored value equals it, on all four transports. Each fails if the handler is put back on `DateTime.UtcNow`.

Validated: all nine unit suites (1,192 core, 276/198/183/242/212/180/36 transports, 222 dashboard), history integration on Memory, SQLite, LiteDB, PostgreSQL, SQL Server and Redis, and a `Release -p:CI=true` build.

## Scope

First item of #304 and does not close it. Still outstanding there: the dashboard stale-message queries, LiteDB's send and heartbeat paths, and the monitors and job scheduler.

## Summary by CodeRabbit

- **Bug Fixes**
  - Message history timestamps now use the queue’s configured time provider across supported transports.
  - Improved consistency between history timestamps, queue timestamps, and duration calculations.
  - Prevented negative durations caused by differences between machine and queue clocks.

- **Breaking Changes**
  - Directly constructed or subclassed history handlers must now provide a time-provider factory. Queues created through the container are unaffected.

- **Tests**
  - Added coverage verifying configured provider timestamps across memory, LiteDB, Redis, and relational database transports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ﻿### Unreleased
+- Fix: message history timestamps come from the queue's configured time provider rather than the machine clock. On SQL Server and PostgreSQL that is the database server's clock, so history agrees with the rest of the queue's timestamps and durations cannot read negative (GitHub #304)
 - ⚠️ Fix: two failures arriving at once each added an error row, so a message got more retries than configured and could loop instead of reaching the error queue. Re-create an existing SQL Server, PostgreSQL or SQLite queue to pick up the fix (GitHub #299)
 - Redis consumers no longer hold a thread through the de-queue. Under thread-pool pressure, where a consumer previously timed out waiting on Redis, it now keeps working (GitHub #256)
 - ⚠️ Custom transports and consumers must add async members: `IReceiveMessages.ReceiveMessageAsync`, `IQueueWait.WaitAsync`, and `IRedisQueueWorkSub.WaitAsync` for Redis. Built-in transports are unaffected (GitHub #256)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ﻿### Unreleased
 - Fix: message history timestamps come from the queue's configured time provider rather than the machine clock. On SQL Server and PostgreSQL that is the database server's clock, so history agrees with the rest of the queue's timestamps and durations cannot read negative (GitHub #304)
+- ⚠️ `WriteMessageHistoryHandler` constructors take an `IGetTimeFactory` on every transport. Only affects code that constructs or subclasses these directly; queues built through the container are unaffected (GitHub #304)
 - ⚠️ Fix: two failures arriving at once each added an error row, so a message got more retries than configured and could loop instead of reaching the error queue. Re-create an existing SQL Server, PostgreSQL or SQLite queue to pick up the fix (GitHub #299)
 - Redis consumers no longer hold a thread through the de-queue. Under thread-pool pressure, where a consumer previously timed out waiting on Redis, it now keeps working (GitHub #256)
 - ⚠️ Custom transports and consumers must add async members: `IReceiveMessages.ReceiveMessageAsync`, `IQueueWait.WaitAsync`, and `IRedisQueueWorkSub.WaitAsync` for Redis. Built-in transports are unaffected (GitHub #256)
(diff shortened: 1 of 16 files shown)


blehnen:
Pushed. Sonar’s 8 and qodo’s 3, plus the Jenkins failure.

**Sonar (8) — both mine.** A duplicate `using System;` in four test files (a scripted edit whose guard only matched the directive when it was not the first line), and S125 on the new field comment, whose second line ended in a semicolon and so read as commented-out code. Neither showed up locally because the test projects are not in `NoTests.sln` and therefore never get the Release build that treats warnings as errors.

**qodo 1 — no end-to-end coverage of the wiring.** Correct, and it is the same gap that hid the PostgreSQL bug in #307: the unit tests construct the handlers directly, so the container could stop handing them the configured provider and they would still pass. `HistoryTimeProviderTest` now registers a clock fixed in 2001 through the ordinary caller override, resolves the handler the queue resolves, writes a record and reads it back through the dashboard’s query — on all six transports. It fails if a handler goes back to `DateTime.UtcNow`.

It found two things on its first run:

- the override has to be on **`IGetTimeFactory`, not `IGetTime`**. Redis registers a factory of its own that answers with the Redis server’s clock and never consults an `IGetTime` registration, so overriding the narrower service left Redis on the real clock.
- asserting the value exposed **#311**: PostgreSQL and SQLite return a history timestamp shifted by the machine’s UTC offset *at that date*, so a column named `EnqueuedUtc` holds local time. Pre-existing — the old code wrote `DateTime.UtcNow`, same `Kind`, same conversion — and out of scope here. The assertion is therefore on which clock the value came from (a quarter century apart), not its exact value, and points at #311.

**qodo 2 — licence headers.** Declined. Test projects are exempt: `CLAUDE.md` → Conventions, *"do not add headers to test files to satisfy this rule, and do not flag their absence."*

**qodo 3 — constructors are a breaking change.** Correct that it breaks direct callers, and intentional. A compatibility overload would have to supply *some* clock, and the only one available without a container is local time — which is the defect this PR fixes, wearing a different name. A compile error is the honest way to tell a direct caller it now has to supply a clock. Changelog carries the ⚠️ note.

---

**On the Jenkins failure** (`PostgreSQL ConsumerErrorTable Run(1,100,5,False,True)`): that DataRow has `enableChaos = true`, and the `40001` / `55000` / `53200` exceptions in the log are **manufactured** — `AddChaosFault` picks a random SqlState from `RetryableSqlErrors`. PostgreSQL raised none of them.

The failure itself is not a slow assertion. The handler must run three times to set `waitForFinish`; the wait is `timeOut * 2` = 200s under chaos, and Jenkins reports the case at 3m53s ≈ 233s, so the full wait elapsed — the third delivery never arrived after one rollback. The queue is also disposed *before* the 30s metric poll, so that poll can never recover a missing rollback.

Not reproduced: 6 clean runs of that class locally against live PostgreSQL. Master is green on the same code. I have not established whether it is chaos flakiness or a real intermittent stall, and with the `retry 1` Jenkins step not working there is no second sample. Happy to chase it under a separate issue rather than hold this PR.

PR #312, opened 2026-09-12
The rest of the timestamps that bypassed the time provider (#304)

blehnen:
Items 2 to 4 of #304. The history handlers were #310; this is the rest, and it closes the issue.

`IGetTime` is registered per transport — the database server's clock on SQL Server and PostgreSQL, the Redis server's on Redis — and code calling `DateTime.UtcNow` silently opts out of whatever the user chose.

## What changed

| | why it matters |
|---|---|
| dashboard stale-message threshold (relational, Redis, LiteDB) | staleness is the age of a heartbeat the *transport* wrote; a cut-off from the asking machine measures the gap between two clocks as well as the age |
| history retention cut-off (`ClearHistoryMonitor`) | compared directly against stored timestamps, so a clock difference moves the retention window by that difference |
| `BaseMonitor.LastRunUtc` | shown on the dashboard beside timestamps the transport wrote |
| Memory `DataStorage.QueuedDateTime` | stored, then compared against later |
| LiteDB: send ×3, receive, heartbeat, error-move, the three find/reset queries | embedded, so the provider is normally the local clock anyway — but these values are compared against each other, and bypassing the provider is how they stop agreeing |

Redis computed its threshold through `DateTimeOffset.UtcNow`, which is why a search for `DateTime.UtcNow` alone did not list it.

## Where to look

Two places keep `DateTime.UtcNow` deliberately, and are the only judgement calls here:

- `JobScheduler` logs the scheduler's time from the provider and the local time on the next line. That is a deliberate side-by-side comparison; converting it would print the same value twice and say nothing.
- The trace decorator and the Dashboard API/UI services measure spans and request timing rather than queue data — where #304 left them.

## Tests

The unit tests for the monitors, `DataStorage` and the three dashboard handlers constructed these types directly, so they now run on a fixed clock; the relational one asserts `@ThresholdTicks` is derived from it rather than from the machine.

Validated locally: `Release -p:CI=true`; unit suites 1192 / 277 / 198 / 183 / 242 / 36 / 222; **LiteDB integration 115/115**, which is the transport this touches most. Memory and Dashboard API integration are left to Jenkins.

## Summary by CodeRabbit

* **Bug Fixes**
  * Queue timestamps now consistently use the configured time provider instead of the machine clock.
  * Improved consistency for message queuing, expiration, heartbeats, stale-message detection, monitor timestamps, and history retention across transports.

* **Breaking Changes**
  * Applications that directly construct or subclass monitors, data storage, or transport handlers must now provide a configured time-provider factory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ﻿### Unreleased
+- Fix: the dashboard's stale-message threshold, the history retention cut-off and LiteDB's send, receive and heartbeat timestamps come from the queue's configured time provider rather than the machine clock, so each is compared against data written on the same clock (GitHub #304)
+- ⚠️ Monitors, `DataStorage` and LiteDB's command and query handlers take an `IGetTimeFactory`. Only affects code that constructs or subclasses these directly (GitHub #304)
 - Fix: message history timestamps come from the queue's configured time provider rather than the machine clock. On SQL Server and PostgreSQL that is the database server's clock, so history agrees with the rest of the queue's timestamps and durations cannot read negative (GitHub #304)
 - ⚠️ `WriteMessageHistoryHandler` constructors take an `IGetTimeFactory` on every transport. Only affects code that constructs or subclasses these directly; queues built through the container are unaffected (GitHub #304)
 - ⚠️ Fix: two failures arriving at once each added an error row, so a message got more retries than configured and could loop instead of reaching the error queue. Re-create an existing SQL Server, PostgreSQL or SQLite queue to pick up the fix (GitHub #299)
--- a/Source/DotNetWorkQueue.Tests/Queue/BaseMonitorTests.cs
+++ b/Source/DotNetWorkQueue.Tests/Queue/BaseMonitorTests.cs
@@ -15,6 +15,9 @@ namespace DotNetWorkQueue.Tests.Queue
     [TestClass]
     public class BaseMonitorTests
     {
+        /// <summary>A date nowhere near the machine clock, so a value from the wrong source cannot match.</summary>
+        internal static readonly DateTime FixedNow = new DateTime(2001, 2, 3, 4, 5, 6, DateTimeKind.Utc);
+
         [TestMethod]
         public void IsDisposed_False_By_Default()
         {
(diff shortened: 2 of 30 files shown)


blehnen:
Pushed.

**Guard ordering (qodo 3 / CodeRabbit ×2) — valid.** Every constructor given an `IGetTimeFactory` called `Create()` before `Guard.NotNull` ran, so a null factory raised `NullReferenceException` where the rest of the library raises `ArgumentNullException`. Fixed in the 13 constructors this branch added it to, and in the 4 from #310 that shipped with the same ordering.

**Expired messages remain (qodo 1, High) — does not reproduce.** I probed LiteDB 5 with the exact shape in the handler — an interface-typed field on `this`, called inside `Where`:

```csharp
private readonly IGetTime _getTime = new FixedTime();   // fixed at 2030
col.Insert(new Row { ExpirationTime = new DateTime(2029, 1, 1) });   // NOT expired by the machine clock
col.Query().Where(x => x.ExpirationTime < _getTime.GetCurrentUtcDate()).ToList();   // returns the row
```

LiteDB translates the call and uses the provider’s value — the 2029 row comes back against the 2030 clock, which the machine clock would not have matched. So expiration cleanup neither throws nor silently stops. That also matches the 115/115 LiteDB integration run, which includes `ConsumerExpiredMessage`.

Hoisted it into a local anyway: one reading for the whole sweep is what is meant, and it matches the sibling handlers. Not presented as a bug fix.

**Missing assertions (qodo 2) — valid, and partly addressed.** Two changed values had a fixed clock injected with nothing asserting it, which is a test that cannot fail for the reason it exists:

- `BaseMonitor.LastRunUtc` — now asserted
- `ClearHistoryMonitor`’s retention cut-off — its first test at all, and the one that matters most here, since that cut-off is compared directly against stored timestamps: a wrong clock moves the retention window rather than merely showing an odd date

Both fail if the production code goes back to `DateTime.UtcNow`. The relational stale cut-off was already asserted via `@ThresholdTicks` in the first commit.

Still not asserted, and I would rather say so than add a test that only re-checks construction: the **LiteDB and Redis stale cut-offs** and the **Memory transport’s `QueuedDateTime`**. Each needs a real database, a Redis server, or the Memory dashboard query to observe, so the unit projects can only cover construction. The honest home is an integration test in the shape of `HistoryTimeProviderTest` from #310 — happy to add it here or as a follow-up, your call.

Validated: `Release -p:CI=true`; unit suites 1194 / 277 / 212 / 180 / 36 / 198 / 183 / 242 / 222.

blehnen:
Filed as #313 rather than adding it here.

The gap is real but it is a different piece of work from this PR: asserting a stale cut-off needs a message held in `Processing` with a live heartbeat across six transports, which is new shared machinery, and it would cover the pre-existing stale path as much as the change here. The risk this PR carries is also narrower than #299 — each change is the same one-line substitution with no per-transport logic, and the relational variant is asserted.

#313 records the non-obvious part so it stays a small task: a clock fixed in the **past** cannot distinguish the two sources for a stale query, because nothing is stale under either. It has to be fixed in the **future** — at 2030 with a 120s threshold a heartbeat written moments ago is older than the cut-off and comes back, while under the machine clock it is not and nothing comes back. It also carries the #310 lesson that the override must be on `IGetTimeFactory`, not `IGetTime`, or Redis silently keeps its own clock.

Issue #303, opened 2026-09-11
Heartbeat scheduler delivers ticks late enough to lose a message's claim


blehnen:
Split out of #294, which has the full history. #302 handles a lapsed claim gracefully; this is the issue about stopping the lapse happening.

## Measurement

The heartbeat on SQLite, instrumented at the point the scheduler delivers a tick and either side of the transport call. Three `ConsumerHeartbeat` runs, 376 ticks:

| | |
|---|---|
| beat duration (send start → send done) | p50 **39ms**, p90 256ms, p99 1.32s, worst 1.53s |
| beats taking longer than the 3s slot | **0** |
| ticks that were skipped because a beat was in flight | **0** |
| tick arrival interval (the cron says 3.00s) | p50 2.97s, p90 3.00s, **worst 9.00s** |

Every tick produced a send. The worker does its work in 39 milliseconds. What varies is when the tick shows up: one arrived **9.00 seconds** after the previous one, against a **10 second** expiry. Two slots vanished, leaving one second before the monitor would have reset a message being actively processed.

That is the lapse mechanism, and it is upstream of `HeartBeatWorker` entirely — nothing the worker or the transport does can widen that margin.

## Why the scheduler is late

`HeartBeatScheduler.AddUpdateJob` schedules onto an **internal memory queue**, and a consumer for that queue executes the beats, with `HeartBeat.ThreadPoolConfiguration.ThreadsMax` bounding it (2 in the integration tests). So every heartbeat in the product is a message produced to a queue, dequeued by a consumer and run through the full processing pipeline — to perform a 39ms update.

That is a lot of machinery between a timer and a row update, and it is shared with whatever else the scheduler is doing.

## Worth considering

- Heartbeats are simple periodic work against a known message. A timer per worker would remove the queue, the consumer, the pool and the scheduling latency in one go.
- If the queue-based scheduler stays, the margin needs to be defensible: with a 3s schedule against a 10s expiry there are three chances, and this measurement shows two can disappear at once.
- `HeartBeatConfiguration` defaults `Time` and `MonitorTime` to `TimeSpan.Zero`, so every caller picks its own ratio and nothing warns that a 3:10 ratio leaves no room for a late tick. Whatever else changes, validating or documenting that relationship is cheap.

## Impact

All transports with a heartbeat — SQL Server, PostgreSQL, SQLite, Redis, LiteDB. The memory transport is exempt because heartbeats are disabled there. SQLite is only where it was easiest to observe.

[...]

PR #315, opened 2026-09-12
Drive heartbeats from a timer instead of the job scheduler (#303)

blehnen:
A worker's heartbeat missed whole slots, and against a 10 second claim that is how a message still being processed is reset by the monitor and handed to a second worker.

Measured on SQLite's `ConsumerHeartbeat`, per message, with a 3 second schedule:

| interval | 3s | 5s | **6s** | **9s** | **12s** |
|---|---|---|---|---|---|
| before | 206 | 12 | **47** | **4** | **1** |
| after | 310 | – | **0** | **0** | **0** |

18.4% of intervals lost at least one slot; none do now. Worst interval 11.99s → 4.23s, p99 9.00s → 3.41s.

## Where the slot was going

Not the plumbing, which is what the issue originally assumed. Dispatch tracked the schedule to 2ms, the scheduler pump never lost a second across 292 iterations, the execution lock was taken 605/605 times across two runs, and the pool was not starved.

The schedule itself skipped: `ScheduledJob` computes the next occurrence from the clock **after the work completes**, so any cycle longer than one period silently drops a beat and nothing counts it.

## What changed

A heartbeat does not need cron — it needs "every N seconds against a deadline". Each worker now owns a `PeriodicTimer` and beats in a serial loop.

`PeriodicTimer` is the specific reason this fixes it: when a tick is missed because the previous beat ran long, the next `WaitForNextTickAsync` returns **immediately** rather than waiting for the following boundary. It fires late; it does not skip.

## Settings kept, with honest meanings

| | |
|---|---|
| `ThreadsMax` | caps concurrent heartbeat **updates**, not threads. Since #284 a beat is an awaited transport call, so this bounds outstanding operations. A beat that cannot get a slot **waits** — never dropped, since dropping is the defect — and a wait longer than the interval is logged. Unset means one per worker, the natural ceiling. |
| `WaitForThreadPoolToFinish` | how long teardown waits for a beat in flight |
| `Time`, `MonitorTime`, `Enabled` | unchanged |
| the four behaviours #302 added to the worker | unchanged |

The transports no longer pin `ThreadsMax` to 1; under the new semantics that would serialise every worker's beat behind a single slot, which with a measured p99 beat of 1.32s is the old bug in a new place.

## Breaking changes

- `HeartBeat.UpdateTime` is a `TimeSpan`: `"*/10 * * * * *"` → `TimeSpan.FromSeconds(10)`
- `IHeartBeatScheduler` and `HeartBeatScheduler` are deleted
- a consumer refuses to start when `Time < UpdateTime * 3`. The worker is only *asked* to stop when its claim lapses and handling code may ignore the token, so refusing the configuration is the only lever the queue has. The error names both values, the attempts remaining, and both ways out.

## What this removes

A job scheduler and Memory queue per consumer with their producer and consumer, a task scheduler, six handlers logging about that carrier queue, and `PendingEvent.Run`'s fire-and-forget `Task.Run` whose faults nothing observed. The job scheduler itself is untouched — it remains a real feature for user jobs.

## Where to look

`HeartBeatWorker.RunLoopAsync` and the teardown around it. `StopLoop` deliberately runs **outside** the beat lock: the loop takes that same lock, so waiting for it while holding it would deadlock.

## Tests

The existing tests asserted lifecycle and queue counts, never the schedule — which is why both this and #301 shipped unnoticed.

- `ABeatThatOverrunsTheInterval_DoesNotForfeitTheSchedule` — fails against a loop that gives up after a slow beat. It does **not** detect a non-serial loop; `BeatOnceAsync`'s own `Running` guard covers that, and the test is named and commented for what it actually holds.
- `HeartBeatGateTests` — the bound, that a waiting beat gets the released slot, and that a slot cannot be released twice
- `ConsumerQueueConfigurationGuardTests` — each rejected ratio, plus the shipped transport defaults, which must keep starting

#302's catch-up flag and its test are gone: a tick cannot arrive during a beat when the loop does not ask for the next one until the beat returns. The guarantee it protected is re-expressed above.

Validated: `Release -p:CI=true`; unit suites 1205 / 277 / 198 / 183 / 242 / 212 / 180 / 36 / 222; heartbeat integration on SQLite, LiteDB and PostgreSQL.

## Summary by CodeRabbit

* **New Features**
  * Heartbeat intervals now use fixed time spans instead of cron expressions.
  * Heartbeat processing uses per-worker timing and configurable concurrency limits.
  * Missed heartbeat intervals are handled after an active update completes.
  * Heartbeat updates continue without indefinite blocking when concurrency limits are reached.

* **Breaking Changes**
  * Consumers must configure `UpdateTime` as a time span.
  * Heartbeat timing must allow at least three update attempts before claim expiration.
  * The scheduler-based heartbeat interface and implementation have been removed.

* **Bug Fixes**
  * Invalid heartbeat settings are now rejected during startup without preventing corrected retries.


blehnen:
Pushed.

**Sonar — one real defect, two taken.** `_loopCancel` was created and never disposed (S2930, blocker). It is disposed in `ReleaseCancel` rather than when the loop is signalled: the loop is awaiting on that token, and disposing it while the wait unwinds races the cancellation. The two drain waits pass `CancellationToken.None` explicitly (S8949 ×2).

**The gate could cost a worker its claim (qodo 1) — correct, and it is this PR's own defect wearing a different hat.** A small `ThreadsMax` serialises beats, and a beat that waited past `HeartBeat.Time` would find its claim lapsed and cancel the worker — over a queue the gate itself created. `EnterAsync` now takes a maximum wait, bounded by the interval; a beat that does not get a slot in time proceeds without one and logs it. The bound is a throttle, not a gate on correctness: a moment of extra database load is cheaper than a lost claim.

**Teardown could hang past its own timeout (qodo 2 / CodeRabbit) — correct.** `StopLoop` bounded the loop wait, then `Dispose`/`DisposeAsync` waited on `_beatLock` with no bound, so a transport call that never returns blocked the message from completing. Both waits share the drain window now and say so when they give up.

**Also fixed:** the guard rejects an interval `PeriodicTimer` cannot accept (qodo 5), rather than letting the consumer start and the first message throw while its heartbeat worker is built; and the synchronous consumer validates *before* setting `Started` (qodo 4), so a rejected configuration can be corrected and `Start` called again — matching what the async one already did.

**Changelog (qodo 6, 7):** the measured figure and the account of the test are gone; the entry carrying both an API removal and a semantics change is now two entries.

**Declined — qodo 3, DotNetWorkQueue.Samples.** Being upgraded when 0.12.0 ships.

---

**On the dashboard failures.** Real, and mine: two test helpers set `HeartBeat.Time` without touching `UpdateTime`, leaving the transport default of two minutes against a five minute claim — exactly the ratio the new guard rejects, and fair to reject. Both now set an interval to match. The short-heartbeat helper needs no beat to land at all, so its interval is set far longer than the test runs rather than left to default. Dashboard integration is **219/219**, twice.

One of those runs earlier showed a single failure in `PurgeHistory_WithDateFilter_Removes_Records`, which turned out not to be this branch: that test only passes as part of the full suite. Filtered to itself it fails on `6f6014c3`, `c80b947d` and `c830e1b1` alike, including a Jenkins-green master. Filed as #316 — it has a `[TestInitialize]` so it looks self-contained, but the Memory transport's store is a process-wide static.

**On coverage.** The gap was where you said — the paths that only run when something goes wrong. Now covered: the gate's timeout, cancellation and post-disposal release (including that a timed-out beat does not release a slot it never took), and teardown when a beat never returns, on both the sync and async paths.

Validated: `Release -p:CI=true`; unit suites 1212 / 277 / 198 / 183 / 242 / 212 / 180 / 36 / 222; dashboard integration 219/219.

blehnen:
Pushed. All three taken.

**Teardown could take two drain windows (Major).** Correct. `StopLoop` waited a full window for the loop, then the beat lock wait started another, so a transport call that never returns cost twice what was configured. Both waits now share a deadline taken before the loop is signalled.

The test covering this was too loose to have caught it — it asserted teardown finished inside fifteen seconds against a 250ms window. It measures against the configured window now: with each wait given its own window it reports 1.01s against a 0.9s bound; with the deadline shared, 0.5s.

**`PeriodicTimer` upper bound.** Correct — `PeriodicTimer` accepts anything under `uint.MaxValue` milliseconds, so the guard was refusing valid intervals such as thirty days for no reason. Corrected to the real bound.

**Stale tests and the heartbeat interval.** The underlying point is right: a beat landing mid-test makes the message look healthy again and the stale query then finds nothing. Fixed at the source rather than by polling around it — the interval is twenty minutes against an hour-long claim, so no beat lands during a test at all. The dashboard decides what is stale from its own `thresholdSeconds`, so the longer claim does not weaken what these tests check, and it removes the race rather than waiting for it to settle.

Validated: `Release -p:CI=true`; unit suites 1212 / 277 / 198 / 183 / 242 / 212 / 180 / 36 / 222; dashboard integration 219/219.

blehnen:
Patch coverage: worth addressing in part, and now done. I measured which lines were actually missing rather than chase the number.

**`HeartBeatGate` 88.2% → 100%.** The uncovered branch was the one reporting a slot that took longer than the update interval — a real condition, and the one that used to be invisible under the old scheduler, where a beat that could not get a thread simply sat in a queue.

**`HeartBeatWorker` 89.2% → 92.1%.** The most useful find was that the substituted logger reported every level as disabled, so the trace and debug branches never ran at all. One of them reads `status.MessageId.Id.Value` — the sort of line that only throws once something reaches it. Also covered: a beat after `Stop` not reaching the transport, and the guard that stops a second beat overlapping one in flight.

**What I left uncovered, deliberately:** four defensive `catch` blocks — a faulted loop during shutdown, a token source disposed underneath the signal, and the loop's outer handler for a failure inside `BeatOnceAsync`'s own failure handling. Each needs a state faked rather than produced, and a test that manufactures it asserts the fake rather than the behaviour. That is most of the remaining 25 lines.

Unit suites: 1215 core (up from 1212), rest unchanged; `Release -p:CI=true` clean.

blehnen:
Taken — the warning could have come from taking the first slot rather than from waiting behind it, since `EnterAsync` measures every acquisition. Calls are cleared once that slot is held, and the queued acquisition is now asserted to be waiting before the delay.

Confirmed it detects what it claims: with the warning branch disabled the test fails; before this change it could pass on a scheduling pause during the first acquisition.

On patch coverage — 88% is the practical limit. What remains is four defensive `catch` blocks in `HeartBeatWorker`: a faulted loop during shutdown, a token source disposed underneath the signal, and the loop's outer handler for a failure inside `BeatOnceAsync`'s own failure handling. Reaching them means faking a state rather than producing one, and the test would then assert the fake. `HeartBeatGate` is at 100%.

Issue #318, opened 2026-09-12
LiteDB: concurrent queue creation can fail with "Member X not found on BsonMapper"


blehnen:
Seen once on Jenkins (PR-317, which touches no LiteDB code), and reproducible on demand.

```
System.NotSupportedException: Member QueueId not found on BsonMapper for type
  DotNetWorkQueue.Transport.LiteDb.Schema.MetaDataErrorsTable.
   at LiteDB.LinqExpressionVisitor.ResolveMember(MemberInfo member)
   ...
   at LiteDB.LiteCollection`1.EnsureIndex[K](Expression`1 keySelector, Boolean unique)
   at ...Schema.MetaDataErrorsTable.Create(...)
```

## Cause

`GetCollection<T>()` uses `BsonMapper.Global`, a process-wide static — the transport never supplies its own mapper. When several threads first touch the same type at once, `EnsureIndex`'s LINQ resolver can see an entity mapper whose member list is not yet populated, and reports a member that plainly exists as missing.

The integration tests run with `Workers: 4, Scope: MethodLevel`, and each test creates a queue, so several `Create` calls race on first use. That is why it appears once in a blue moon rather than never or always.

## Reproduction

64 concurrent first-uses of one type, in a fresh process:

```csharp
Parallel.For(0, 64, i =>
{
    var col = db.GetCollection<Row>("c" + i);
    col.EnsureIndex(x => x.Id);
    col.EnsureIndex(x => x.QueueId, true);
});
```

Fails every run, with the same message shape.

## What fixes it

Naming the field instead of passing a lambda keeps LiteDB's LINQ resolver — the racy part — out of the path:

```csharp
col.EnsureIndex("QueueId", "$.QueueId", true);
```

**0 failures across 4 runs of 64**, against a failure on every run with the lambda form.

Pre-warming the mapper first (`BsonMapper.Global.ToDocument(new Row())`) is **not** sufficient: it reduces the failures but one run still produced 1 of 64, so something on the `EnsureIndex` path builds state the warm-up does not.

## Scope

21 lambda `EnsureIndex` calls across the LiteDB schema classes — `ConfigurationTable`, `ErrorTrackingTable`, `MetaDataTable`, `MetaDataErrorsTable`, `JobsTable`, `StatusTable`. All are index creation at queue-create time, so switching them changes no stored data.

Worth checking while there: whether anything else in the transport resolves members through LINQ on first use.

blehnen:
[...]

PR #319, opened 2026-09-12
Serialise LiteDb table creation (#318)

blehnen:
Creating two queues at once could fail:

```
NotSupportedException: Member QueueId not found on BsonMapper for type
  DotNetWorkQueue.Transport.LiteDb.Schema.MetaDataErrorsTable
   at LiteDB.LinqExpressionVisitor.ResolveMember(MemberInfo member)
   at LiteDB.LiteCollection`1.EnsureIndex[K](Expression`1 keySelector, Boolean unique)
```

Seen on CI on a branch touching no LiteDb code, then on master.

## Cause

`GetCollection<T>` uses `BsonMapper.Global`, a process-wide static, and mapping a type for the first time is not thread safe: a second thread can see an entity mapper whose member list is not yet populated, and `EnsureIndex`'s LINQ resolver then reports a member that plainly exists as missing. The integration tests run four workers and every test creates a queue.

## Fix

Table creation takes a lock. Sixty-four concurrent first uses of one type reproduce the fault on **every** run; with the lock, **zero failures over four runs of sixty-four**. Creating a queue is a rare, one-off operation, so serializing it costs little.

## Two alternatives measured and rejected

- **Pre-building the mappers** (`BsonMapper.Global.ToDocument(...)`) still failed **one run in sixty-four** — something on the `EnsureIndex` path builds state the warm-up does not. This was the obvious fix and it does not hold.
- **Naming the field** — `EnsureIndex("QueueId", "$.QueueId", true)` — does avoid the resolver and gave zero failures, but it is twenty-one call sites and the two forms are not interchangeable: `EnsureIndex(x => x.Id)` resolves to the built-in `$._id` index rather than `$.Id`, returns `false`, and does not create the collection. A mechanical swap would have written an index on a field that does not exist.

## The test, and what it cannot do

`ConcurrentQueueCreation` creates sixteen queues at once. It **cannot** reproduce the original fault, and says so in its summary: that needs the first use of a type in the process, and any earlier test in the assembly warms the mapper. It is here because concurrent creation is worth exercising — a deadlock or a file-level collision in this path would fail it — and the serialisation it guards had no test at all. The reproduction that does bite is recorded in #318.

Validated: `Release -p:CI=true`; LiteDb unit 180/180, LiteDb integration 116/116.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when creating multiple queues concurrently.
  - Prevented intermittent setup failures caused by simultaneous database table initialization.
  - Ensured queue and job database structures are initialized safely when created at the same time.

- **Tests**
  - Added coverage verifying that multiple queues can be created concurrently and successfully.
  - Added safeguards to detect stalled concurrent creation during testing.

### DIFF
--- a/Source/DotNetWorkQueue.Transport.LiteDB.IntegrationTests/ConcurrentQueueCreation.cs
+++ b/Source/DotNetWorkQueue.Transport.LiteDB.IntegrationTests/ConcurrentQueueCreation.cs
@@ -1,0 +1,72 @@
+using System;
+using System.Collections.Concurrent;
+using System.Threading.Tasks;
+using DotNetWorkQueue.Configuration;
+using DotNetWorkQueue.IntegrationTests.Shared;
+using DotNetWorkQueue.Transport.LiteDb.Basic;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace DotNetWorkQueue.Transport.LiteDb.IntegrationTests
+{
+    /// <summary>
+    /// Several queues being created at once.
+    ///
+    /// LiteDB maps a type through a process-wide BsonMapper, and mapping one for the first time is not
+    /// thread safe - a second thread can see a half-built mapper and report a member that exists as
+    /// missing. Creating tables is serialised because of it (GitHub #318).
+    ///
+    /// What this test can and cannot do is worth being plain about: the fault needs the *first* use of
+    /// a type in the process, so once any other test in this assembly has touched LiteDB the mappers
+    /// are warm and it will not reproduce here whatever the lock does. It is kept because creating
+    /// queues concurrently is worth exercising on its own - a deadlock or a file-level collision in
+    /// this path would fail it - and because the serialisation it covers has no other test.
+    /// </summary>
+    [TestClass]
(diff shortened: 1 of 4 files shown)


blehnen:
Pushed.

**Job table creation was not covered — correct, and it was the real gap.** `CreateJobTablesCommandHandler` creates `JobsTable` through the same global mapper without the lock, so a job queue starting alongside a message queue raced exactly as two message queues did. Brian spotted the same thing.

The gate now lives in its own type and both handlers take it. A gate per handler protects nothing here: the race is on one process-wide `BsonMapper`, so the lock has to be one object for every path that maps a type for the first time.

**Timeout on the concurrency test — taken.** It exercises a lock, so a mistake there hangs the run rather than failing it, and the repository already bounds the seven tests that can block.

**Licence header on the new test file — declined.** Test projects are exempt: `CLAUDE.md` → Conventions, *"Test projects are exempt and mostly do not carry one… do not add headers to test files to satisfy this rule, and do not flag their absence."*

Validated: `Release -p:CI=true`; LiteDb unit 180/180, LiteDb integration 116/116.

Issue #352, opened 2026-09-16
Rollback drops the queue delay when a heartbeat claim is present, starving the queue


blehnen:
## What happens

With `EnableDelayedProcessing` on, a consumer that rolls a message back re-reads that same message forever and never reaches the messages behind it. One repeatedly-cancelled message is enough to starve the whole queue.

Affects **PostgreSQL, SQLite and SQL Server**. LiteDb and Redis are not affected.

## Why

The de-queue orders by `QueueProcessTime`, and the rollback is what moves that value forward so a rolled-back message goes to the back of the line.

`RollbackMessageCommandHandler.PrepareRollback` builds its statement from two flags. On the branch that carries a heartbeat it passes `false` for the first one, which is `includeDateIfEnabled`:

```csharp
if (_options.Value.EnableDelayedProcessing && rollBackCommand.IncreaseQueueDelay.HasValue)
{
    if (rollBackCommand.LastHeartBeat.HasValue)
        command.CommandText = GetRollbackSql(false, true);   // QueueProcessTime not in the update
    else
        command.CommandText = GetRollbackSql(true, false);
    ...
    command.Parameters["@QueueProcessTime"].Value = dtUtcDate;   // bound, but the SQL does not use it
}
```

So `QueueProcessTime = @QueueProcessTime` is left out of the UPDATE while the parameter is still bound. Nothing errors and nothing is logged; the message simply keeps its original enqueue time and stays first in line.

`SetupSql` pre-generates four variants including `"TrueTrue"`, which no caller ever asks for. That was the tell.

## Why it only just started

The branch was previously reached only once a worker's first beat had landed. Since #336 the de-queue publishes its claim immediately, so `LastHeartBeat` is always set and **every** rollback now takes it.

## Impact

Measured against PostgreSQL: a consumer rejecting messages spun at ~190 de-queue/rollback cycles per second on the same two rows, and three of seven messages were never reached at all.

`ConsumerMethodCancelWork` fails 7 times in 8 on master and 0 times in 7 immediately before #336.

## Fix

Ask for `GetRollbackSql(true, true)` on that branch, which also makes all four pre-generated variants reachable.

PR #354, opened 2026-09-16
fix: set the error count rather than incrementing it (#350)

### LINKED ISSUE #350: The error-count upsert is not idempotent, and it is retried
## What

The error-count upsert is not idempotent, and it is wrapped in a retry decorator. A transient fault raised *after* the server committed the statement therefore costs the message a retry it never used.

PostgreSQL (`PostgreSQLCommandStringCache.cs`, `UpsertErrorCount`):

```sql
insert into {ErrorTracking} (QueueID, ExceptionType, RetryCount)
values (@QueueID, @ExceptionType, 1)
on conflict (QueueID, ExceptionType)
do update set retrycount = {ErrorTracking}.retrycount + 1
```

And the retry decorator is registered **open-generic over every command handler** (`PostgreSQLMessageQueueInit.cs`):

```csharp
container.RegisterDecorator(typeof(ICommandHandler<>),
    typeof(RetryCommandHandlerDecorator<>), LifeStyles.Singleton);
```

so `SetErrorCountCommand` is retried along with everything else.

## Why it matters

`retrycount + 1` is a relative update. The retry policy exists for exactly the faults where the outcome is unknown to the client — a deadlock victim, a connection dropped after commit, a timeout waiting for a result that had already been written. In the "committed but not acknowledged" case the retry increments a second time for one real failure.

The message then reaches its configured retry limit early and is moved to the error queue before it has had all its attempts. From the outside that looks like a message that gave up too soon, with nothing in the logs to say why, because both the original write and the retry succeeded.

How often this bites depends on how often those faults occur, which in normal operation is rare. It is not rare on a loaded CI box, and it is deliberately not rare when chaos is enabled.

## What this is not

**Chaos cannot cause it.** Polly's fault strategy is proactive: when it injects, the user callback is not invoked at all, so the statement never reaches the server and the retry runs it exactly once. Checked in the Polly documentation rather than assumed, while investigating #314 — that route is ruled out.

This is about genuine transients only.

## Also affected

The same shape exists on the other relational transports (`UpsertErrorCount` / the check-then-write fallback) and should be checked per transport rather than assumed identical. LiteDb's handler does the increment in memory and writes the record back, which has a different failure mode.

## Possible fixes

1. **Make the write idempotent** — carry an attempt identifier so a replay is recognisable, and set rather than increment. Cleanest, and the most work: it changes what the error table stores.
2. **Exclude this one command from the retry decorator.** A lost increment (no retry attempted, the failure propagates) means the message gets an *extra* attempt rather than one fewer, which is the safer direction — a message that retries once too often eventually reaches the error queue anyway, where one that retries too few is silently short-changed.
3. **Accept it and document it**, on the grounds that the window is narrow. Worth stating explicitly if that is the decision, rather than leaving it as an accident.

Option 2 is the smallest and moves the failure in the safe direction, at the cost of losing retry protection on one command.

## How it was found

Reading the error-count path while investigating #314, which shows a message reaching the error queue after two deliveries rather than three. This mechanism would produce exactly that shape — but chaos being ruled out means it is not the explanation for #314's chaos-only row, and #314 has its own diagnostics now. Filed separately because it stands on its own.

blehnen:
Fixes #350.

The error-count write is wrapped in a retry policy and was a relative update. A transient fault raised *after* the server had already committed it therefore counted one real failure twice, and the message reached its configured retry limit early — moved to the error queue with attempts still owed to it, silently, because both the original write and the replay succeeded.

## What changed

`SetErrorCountCommand<T>` carries the total rather than implying an increment. The caller already has it — `ReceiveErrorMessage` reads the count on the line above — so passing `retries + 1` costs nothing and makes a replay a no-op.

```diff
-values (@QueueID, @ExceptionType, 1)
+values (@QueueID, @ExceptionType, @RetryCount)
 on conflict (QueueID, ExceptionType)
-do update set retrycount = {ErrorTracking}.retrycount + 1
+do update set retrycount = @RetryCount
```

All three of `UpdateErrorCount`, `InsertErrorCount` and `UpsertErrorCount` on PostgreSQL, SQLite and SQL Server; LiteDb sets rather than increments in memory. Redis has no error tracking table.

## The trade, and a test that had to change

This swaps one failure mode for another **on purpose**. Two workers failing the same message at the same instant both read the same count and both write the same total, so one failure goes uncounted and the message gets one attempt more than configured. An extra attempt still ends at the error queue; a missing one does not, and it is silent.

**#299's `CountsConcurrentFirstFailures_AsOneRow` asserted the old behaviour directly** — 32 concurrent failures counting as 32 — so it now expects 1, with the reasoning inline. Its `RowCount == 1` assertion, which is what #299 was actually about, is unchanged. Please look at that hunk specifically.

## What to look at

- `SetErrorCountCommand` — the remarks explain why absolute, and the guard rejects a count below 1 since a total is never zero.
- The changed assertion in the three `SetErrorCountAgainstBothSchemasTests`.
- The idempotence check extends that same test rather than adding a third file, so it covers **both** the atomic upsert and the older check-then-write fallback: write the total, write it again, the count must not move. Reverting the PostgreSQL upsert to `retrycount + 1` fails it with `replaying the same total counted a second failure` (verified).

## Also in here

One changelog line the #353 entry should have carried. The same rollback statement #352 fixed also applies `RetryDelayBehavior`, so the configured back-off was not being applied either — measured at 0.04s against a configured 8s.

## Verification

Error-path suites (`ConsumerErrorTable`, `ConsumerPoisonMessage`, `ConsumerRollBack`, `SetErrorCount`): PostgreSQL 18, SQLite 10, SQL Server 10, LiteDb 8, Memory 4. Unit: core 1226, RelationalDatabase 278, SQL Server 198, PostgreSQL 183, SQLite 248, LiteDb 182. Zero failures. Release build clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Retry counts are now recorded accurately across SQL Server, PostgreSQL, SQLite, and LiteDB, preventing duplicate counting during transient failures or concurrent processing.
  - Retry-count updates preserve the highest recorded total when writes are replayed or arrive out of order.
  - Configured retry delays are now honored, ensuring failed messages wait for the specified back-off period before being retried.


blehnen:
Pushed `089b5780`. All four findings were valid; here is what changed.

**Error counts could move backwards** (qodo High, also raised by CodeRabbit) — real, and the path is specific to this codebase rather than theoretical: a worker reads the count and stalls, its heartbeat claim lapses, another worker takes the message and advances the row, and the first worker's write lands afterwards carrying the old value. The count drops and the message is handed attempts it has already used — the exact failure this PR set out to remove, arriving by another route.

Every update now keeps the greater of stored and supplied. Replay idempotence is untouched, since a replayed total equals the stored one:

| | |
|---|---|
| PostgreSQL | `GREATEST(retrycount, @RetryCount)` |
| SQLite | `max(retrycount, @RetryCount)` |
| SQL Server | `CASE WHEN retrycount > @RetryCount ...` |
| LiteDb | `Math.Max(...)` |

`CASE` rather than `GREATEST` on SQL Server deliberately — `GREATEST` is 2022 and later, and this is not the change to raise the supported floor with.

A staggered-write assertion is now in all four transports: write 2, replay a stale 1, the count must stay 2.

**LiteDb had no test of its own** — correct, and it matters more there than elsewhere because LiteDb does its own read and write instead of handing a statement to a server, so it can drift from the other three silently. New `SetErrorCountTests` asserts the same two properties. Reverting its handler fails it either way: `= command.RetryCount` gives *a stale lower total overwrote a higher one*, `= RetryCount + 1` gives *replaying the same total counted a second failure*.

The first version of that test was wrong in a way worth recording: it opened a second `LiteDatabase` on the file while the container held it exclusively, so it saw none of the writes and was measuring its own plumbing. It reads through the library's own query handler now.

**Nothing covered `retries + 1`** — also correct, and it was the one place a regression would be invisible, since the handler tests are handed a number that is already computed. `ReceiveErrorMessageRetryCountTests` covers a first failure, a later one and the async path; all three fail if the addition is dropped.

That test needed a fix too: `RetryDelayBehavior` comes from a factory, so the configuration AutoFixture produces holds a substitute whose `Add` does nothing and whose `MaxRetries` is zero — every message went straight to the error queue, past the line under test. It would have passed for the wrong reason against a weaker assertion. The retry information is stubbed directly now, with a comment saying why.

**Changelog** — over the length limit, carried storage mechanics and measured timings, all three against the rules in CLAUDE.md. Both entries rewritten to the user-visible outcome; now 274 and 213 characters.

Verification: PostgreSQL 18, SQLite 10, SQL Server 10, LiteDb 9, RelationalDatabase 281, core 1226, SQL Server 198, PostgreSQL 183, SQLite 248, LiteDb 182. Release build clean.

PR #355, opened 2026-09-16
fix: refuse a producer or consumer whose queue does not exist (#348)

### LINKED ISSUE #348: A producer or consumer pointed at a queue that does not exist yet is permanently broken
## What

Pointing a producer, consumer or dashboard at a queue that does not exist yet leaves it permanently broken, in a way that neither fails cleanly nor recovers. #120 describes one part of this — the options factory caching defaults — but fixing that alone does not fix any of the three, for reasons set out below.

Measured against SQLite, with the queue created **after** the container was built, which is the ordering the library has never enforced.

### Producer: fails every send, forever

```
before create: status=True heartbeat=True delayed=False priority=False   <- cached defaults
create:        Success   (real queue has different options)
after create:  status=True heartbeat=True delayed=False priority=False   <- still the defaults
send:          HasError=True  SQLiteException: SQL logic error
```

The stale options drive SQL generation, so the statement references columns the real table does not have. The mirror case — real table has *more* than the defaults — fails as `constraint failed`. Both are loud, neither is legible: nothing in either message points at stale options.

### Consumer: dies silently after one error

```
after 4s with no queue:      errorsRaised=1  handled=0
create: Success              fresh producer send: HasError=False   <- the queue is fine
  +5s … +30s after create:   errorsRaised=1  handled=0
```

It raises **one** error and then stops. The count never climbs, so it is not retrying — and nothing further appears in the logs. A consumer deployed before its schema migration looks alive and processes nothing, indefinitely.

### Dashboard: shows defaults

The case #120 was filed for. Least severe of the three: wrong display rather than broken operation.

## Why fixing the options factory is not enough

**1. There is a second cache under the first.** 28 handler files across SQL Server, PostgreSQL, SQLite and LiteDb wrap the factory in their own `Lazy<T>`:

```csharp
_options = new Lazy<SqlServerMessageQueueTransportOptions>(optionsFactory.Create);
```

Whatever the factory does later, each handler pins the first value it resolved for that handler's lifetime.

**2. The generated SQL is cached, and the key is options-blind.**

```csharp
private static string BuildCacheKey(List<string> routes, string userQuery)
{
    if (!string.IsNullOrEmpty(userQuery)) return null;
    var routeCount = routes?.Count ?? 0;
    return routeCount == 0 ? DequeueKey : $"{DequeueKey}|routes={routeCount}";
}
```

Routes and the user clause only. Statements are cached at **build** time, not on success, so the first attempt against a non-existent queue poisons the cache with statements built from default options. The consumer does not hit this today only because it dies first — **it becomes the next bug the moment retrying is added.**

**3. In-place mutation of the options object does not solve it either.** Nothing freezes the transport options (`SetReadOnly()` is called on the configuration objects, never on these), so mutating one after publication is mechanically possible and would reach every `Lazy<T>` holding the reference. But it would leave the command caches serving the old SQL, so it trades one silent staleness for a subtler one.

## Why the previous attempt was reverted

PR #137 removed the cache on the default fallback so each `Create()` retried the load. That turned one read into N concurrent reads during the creation window — `setOptions`, the CreateQueueTables handler, producer init, consumer init and the heartbeat monitor all resolve back to back. Those reads hit `sqlite_master` / the SQL Server catalog while `CREATE TABLE` was running (`database schema is locked`, `Execution Timeout Expired`), Polly backed off 3.9s → 17s → 21s, and it cascaded into multi-hour Jenkins stalls.

Any fix must not poll the store per resolve.

## The decision this needs first

**Should a producer or consumer self-heal when the queue appears, or refuse to start without it?**

Today it does neither, which is the worst of the three options: it looks alive and does nothing.

Arguments for self-healing, which is the recommendation here:

- The library has **never enforced an ordering constraint** between queue creation and producer/consumer startup. Introducing one now is a breaking change to deployment for every consumer of the library.
- A consumer polls anyway. "Not there yet, keep polling, start when it appears" costs nothing it is not already spending.
- A producer cannot buffer, so each send should fail with a clear error and the *next* should succeed once the queue exists.

**Redis is different, and needs less rather than a special case.** It has no schema to mismatch — a non-existent Redis queue is an empty keyspace, so a consumer finds nothing and waits, which is already the self-healing behaviour. Options staleness still applies if the configuration key is written later, but there is no SQL shape to violate and therefore no hard failure.

## Scope, if self-healing is chosen

1. A "queue does not exist yet" state that is recoverable rather than terminal, in the receive and send paths.
2. Command cache invalidation when the options that generated the statements change — otherwise (2) above bites immediately.
3. Options refresh that does not poll the store per resolve, given #137.
4. A specific exception type for the genuinely-misconfigured case, so the degraded state is legible instead of surfacing as `SQL logic error`. Raised by @blehnen on #120 as a possible whole answer; better as a diagnostic alongside recovery than as a hard failure at construction.

## Supersedes #120

#120 is accurate about the options factory but describes roughly a third of the problem, and its proposed fix (option 1, "don't cache the default fallback") is the one that was tried and reverted.

blehnen:
Fixes #348.

Pointing a producer or consumer at a queue that has not been created neither failed cleanly nor recovered. Measured on current master — one consumer started before the queue existed, then the queue created and a message sent:

| transport | errors raised | messages handled in 30s |
|---|---|---|
| **SQLite** | 1 | **0** |
| **LiteDb** | **0** | **0** |
| SQL Server | 3 | 1 — recovered |
| PostgreSQL | 3 | 1 — recovered |
| Redis | 0 | 1 — recovered |

A fresh consumer against the same queue worked immediately in every case, so on SQLite and LiteDb the damage is per-instance and permanent. LiteDb manages it without a single log line: a consumer that looks healthy and processes nothing for as long as it runs.

## What changed

The check runs where the caller can still do something about it — `QueueContainer`'s producer and consumer creation — and costs **one read per producer or consumer, not one per resolve**, which is what made #137 unusable.

- `QueueDoesNotExistException`, naming the queue, instead of `SQL logic error` or silence.
- `IQueueCreation.RequiresCreation` — the exemption switch.
- `IContainer.TryGetInstance<T>()`, because a transport is not obliged to register `IQueueCreation` (the core tests' own no-op transport does not), and "nothing registered" must not be confused with "registered and failed to build".

**Three deliberate exemptions**, each of which cost a test before I understood it:

- **Redis and Memory.** Redis keeps no schema, `CreateQueue` is a documented no-op, and `QueueExists` answers on *content* — so an empty queue reads as absent. Checking would refuse every correctly configured consumer at the moment it starts.
- **Job producers.** A job producer builds the storage it needs; `JobProducer_ReachingADatabaseFirst_SetsWalMode` exists to prove it works against a database with nothing in it.
- **Admin paths.** The dashboard only calls `CreateAdminContainer`, so it still shows broken queues — which is its job.

## Two bugs the tests flushed out

**LiteDb reported a queue as non-existent immediately after creating it.** It asked about a collection named after the queue, but LiteDB materialises a collection on its first *write* and creation writes configuration only. That also meant `CreateQueue` could not report `AlreadyExists` and `RemoveQueue` refused to remove a queue that was there. Now asks about the configuration collection. The test asserts `QueueExists` directly so this cannot regress behind the guard.

**`BatchSendResults` was passing against an empty database.** It created the queue, disposed the creation container, then built a producer in a fresh one without sharing the creation scope — so for an in-memory database the producer had its own, different one. The send succeeded against nothing and the assertions could not tell. It shares the scope now.

## What to look at

- `QueueContainer.GuardQueueExists` and the seven call sites — in particular which paths are *not* guarded.
- `RequiresCreation` on `RedisQueueCreation`, and the remark explaining why Redis cannot use `QueueExists` for this.
- `QueueMustExistTests` (SQL Server, PostgreSQL, SQLite, LiteDb) and `QueueNeedNotExistTests` (Redis).

## Verification

Release build clean, full Debug build clean. Core 1226/1226; RelationalDatabase 281; SqlServer 198; PostgreSQL 183; SQLite 248; LiteDb 182; Redis 212; Memory 36. Integration: SqlServer 170/170, PostgreSQL 202/202, LiteDb 122/122, Memory 69/69, SQLite all chunks 0 failures, Redis Basic/Producer/Consumer 53/53.

`StarvationReceiveBaselineTests` has one failing test, **pre-existing on master** — verified by running that class on master and on this branch: 1 failed / 1 passed either way, 6m58s vs 6m54s. Filed separately, not touched here.

## Summary by CodeRabbit

* **New Features**
  * Database-backed transports now reject consumers and producers for queues that have not been created, with a clear `QueueDoesNotExistException`.
  * Queue creation status is reported consistently across supported transports.
  * Added a safe way to check whether a service is registered without requiring resolution.
* **Bug Fixes**
  * LiteDB recognizes queues immediately after creation, including empty queues.
  * Redis and in-memory queues remain usable without explicit creation.
* **Tests**
  * Added coverage for queue validation and creation behavior across transports.


blehnen:
Pushed `2d9c7d03`. Four of five actioned, one declined.

**1. Failed creations leave schedulers running** — valid, and worse than described. `CreateConsumerQueueScheduler(queueConnection)` starts a scheduler and retains its container *before* delegating to the method that throws:

```csharp
factory.Scheduler.Start();
Containers.Add(schedulerCreator);
return CreateConsumerQueueSchedulerInternal(...);   // throws in here
```

So a refusal left a **running scheduler** plus two containers owned by the parent for its lifetime, accumulating per attempt. Three changes: the guard runs before `Containers.Add`, it disposes the container it refuses, and both scheduler overloads dispose theirs if the inner call throws.

**2. Custom transports no longer compile** — valid, and you pointed at a precedent I had missed. `IContainer.GetImplementationType<TService>() => null` is already a default member, with the remark *"so an `IContainer` implemented elsewhere is unaffected"*. Both new members are defaulted now (`null` / `false`), which preserves existing behaviour exactly: a transport or container written before this simply is not checked until it opts in. The interface change is no longer breaking, the changelog entry has lost its ⚠️, and the test double that had to grow a member went back to what it was.

**3. Sample applications fail at startup** — not valid. Checked the samples repository directly: **every schema-backed sample already does this by hand.** `SQLServerConsumer/Program.cs` resolves `SqlServerMessageQueueCreation`, checks `QueueExists`, and returns with *"Verify that you have run the producer, which will create the queue"* before building the consumer. Producers call `CreateQueue()`. Across all 30 sample programs the only ones without a guard are the four Redis consumers — and Redis is exempt.

This PR moves that hand-rolled check into the library rather than conflicting with it. The samples could be simplified afterwards, but they are not broken by it.

**4. Memory exemption lacks coverage** — valid. Added `QueueNeedNotExistTests` for Memory. It is exempt for a different reason from Redis (Redis has a store with no schema; Memory has no store), so it is worth its own test rather than leaning on the Redis one. Memory integration 69 → 70.

**5. Missing LGPL headers** — declined. CLAUDE.md, line 127: *"Test projects are exempt and mostly do not carry one; #206 scoped its header sweep to production files deliberately. Do not add headers to test files to satisfy this rule, and do not flag their absence."* All three flagged files are tests.

Verification: Release build clean, core 1226/1226, Memory integration 70/70, SQL Server integration 170/170.

Issue #336, opened 2026-09-14
Heartbeat: the worker never learns the heartbeat the de-queue wrote


blehnen:
## What

The heartbeat a worker holds a claim with is written twice by two different parties, and the worker only ever learns about the second one.

The de-queue stamps the record:

| transport | where |
|---|---|
| SQL Server | `CreateDequeueStatement.cs` — `HeartBeat = GetUTCDate()` |
| PostgreSQL | `ReceiveMessage.cs` — `HeartBeat = @CurrentDate` |
| SQLite | `ReceiveMessage.cs` — `HeartBeat = (select CurrentDateTime from …)` |
| LiteDb | `ReceiveMessageQueryHandler.cs` — `record.HeartBeat = _getTime.GetCurrentUtcDate()` |
| Redis | `DequeueLua.cs` — the working-set score |

None of them return it. `WorkerNotification.HeartBeat.Status` is only populated once the worker's *own* first beat lands, so between the de-queue and that first beat the worker holds a claim whose value it does not know.

## Why it matters

Two paths test ownership against that value, and neither can while it is unknown:

- **The beat** (`SendHeartBeat` → `and (@previous is null or HeartBeat = @previous)`). #335 handles the unknown case by refusing to send once the claim is past the expiry, and cancelling instead. Correct, but it gives up a claim that may well still be ours.
- **The rollback** (`RollbackMessageCommandHandler` → `AND heartbeat = @HeartBeat`). This one has been there since the initial check-in, and when `LastHeartBeat` is null the clause is **omitted entirely** — so a rollback with no beat yet resets the row unconditionally, including a claim that now belongs to another worker. Same hole, opposite handling, and still open.

## What would fix it

Return the de-queued heartbeat with the message and seed `WorkerNotification.HeartBeat.Status` from it, so both paths have a real value to compare from the moment the message is picked up.

Cost is the reason it is not in #335: de-queue SQL, the readers, and the received-message plumbing on all five transports, plus `IHeartBeatStatus` construction moving to receive time. That is a wide change for a stability release.

## Rejected alternative

Comparing "at or before the value we know about" (`<=`) seeded from the worker's own start time. SQL Server stamps the de-queue heartbeat with `GetUTCDate()` — the database's clock, not the worker's — so clock skew would fail every legitimate first beat. It also weakens the reclaim check it is meant to strengthen, since a successor's value only has to be newer by the same clock.

## Notes

[...]

PR #358, opened 2026-09-17
cleanup: SonarCloud sweep before the release (#346)

### LINKED ISSUE #346: SonarCloud cleanup before the next release (56 open: 6 real, 50 from the MSTest 4.4.0 analyzers)
## What

SonarCloud is at **56 open issues** since 0.12.0 shipped. Cleaning them up is the last thing before the next release, the same way #324 was for 0.12.0.

They are not one problem, and the split matters for how they get handled:

| count | rule | severity | what it is |
|---|---|---|---|
| 47 | `MSTEST0068` | INFO | `Use 'Assert.AreSequenceEqual' instead of 'CollectionAssert.AreEqual'` |
| 3 | `MSTEST0037` | INFO | `Use 'Assert.IsEmpty' instead of 'Assert.HasCount'` |
| 4 | `plsql:LiteralsNonPrintableCharactersCheck` | **CRITICAL** | `An illegal character with code point 13 was found in this literal` |
| 2 | `plsql:S1192` | **CRITICAL** | duplicated literal in a SQL script |

## The six CRITICALs are real, and they are in the upgrade scripts

Both plsql rules fire on `docs/upgrade/0.12.0/postgresql.sql` and `sqlserver.sql`, added in #338.

**Code point 13 is a carriage return**, and it is landing *inside multi-line SQL string literals* — the `EXECUTE format('...')` blocks, which span lines. The repository has no `*.sql` rule in `.gitattributes`, so those files fall under `* text=auto` and are checked out CRLF on Windows, which is where the SonarCloud scanner runs (`runs-on: windows-latest`).

Harmless to PostgreSQL, which treats the CR as whitespace inside dynamic SQL. Worth fixing anyway, because these are **operator-facing scripts** run through `psql`, `sqlcmd` and assorted clients on assorted platforms, and a stray CR inside a quoted literal is the kind of thing that behaves differently in one of them.

The fix is probably `*.sql text eol=lf` in `.gitattributes` rather than editing the scripts — SQL here is run on Linux at least as often as Windows. Worth confirming the scanner result changes before assuming that settles it.

`S1192` is the ordinary duplicated-literal rule; needs reading rather than a blanket extract.

## The 50 INFO ones came in with MSTest 4.4.0

Both analyzers arrived with the bump in #341. Unlike the `MSTEST0065` batch handled in #342, **these are not asserting the wrong thing**: `CollectionAssert.AreEqual` does compare elements. It is a newer-API preference, so the work is mechanical — but it is 47 sites, and each should get the same "is identity or contents being asserted here" read that #342 applied, rather than a find-and-replace.

## Ordering

The six CRITICALs are worth doing on their own merit and are small. The 50 INFO ones are the bulk and can wait for the pre-release sweep.

Prior art: #324 (0.12.0's sweep), #342 (the `MSTEST0065` batch), and `docs/lessons-learned.md` for the SonarCloud API details — accept via `POST /api/issues/bulk_change`, and the free plan cannot disable rules, only accept per issue.

blehnen:
Closes #346 once the two accept-passes below are done.

SonarCloud was at **64** open, not the 56 the issue counted — **five of the new ones are mine**, from #355 and #336.

| count | rule | what happened |
|---|---|---|
| 47 | `MSTEST0068` | converted, see below |
| 6 | `MSTEST0037` | `HasCount(0, x)` → `IsEmpty(x)`; `AreEqual(n, x.Count())` → `HasCount(n, x)` |
| 4 | `LiteralsNonPrintableCharacters` **CRITICAL** | `.gitattributes` |
| 3 | `S4136` | mine — overload pairs separated |
| 2 | `S1192` **CRITICAL** | accept |
| 2 | `S107` | accept |

## The one that needed judgement

`MSTEST0068` suggests `Assert.AreSequenceEqual` for **both** `CollectionAssert.AreEqual` *and* `CollectionAssert.AreEquivalent`. Those are not the same assertion — `AreEquivalent` does not care about order.

Twelve sites depend on that: which tables a delete cleared, what a scheduler lists, what a dictionary holds. Taking the suggestion literally would have quietly imposed an ordering none of them mean to assert — passing today, and failing later for a reason that has nothing to do with the thing under test.

`Assert.AreEquivalent` is the actual replacement and exists in 4.4. Being generic, it also removed twelve `(System.Collections.ICollection)` casts.

Four of those twelve needed more than a rename: two where the sides are `IReadOnlyDictionary` and `IDictionary` so there is no `T` to infer, and two where the binding overload takes a span rather than a `List`.

Separately, `CollectionAssert.Contains(collection, item)` and `Assert.Contains(item, collection)` take their arguments in opposite orders, so those nine swapped rather than renamed.

## The CRITICALs

All four are one cause: a carriage return inside the multi-line quoted literals in the 0.12.0 upgrade scripts. The scanner runs on `windows-latest`, and with no `*.sql` rule these fall under `* text=auto` and check out CRLF there.

Harmless to PostgreSQL, which treats it as whitespace inside dynamic SQL — but these are operator-facing scripts run through whatever client an operator has, and a stray CR inside a quoted literal is not handled identically by all of them. Pinned to LF in `.gitattributes` rather than editing the scripts, since they are run on Linux at least as often as Windows.

**This one needs the rescan to confirm** — the fix is to what the scanner checks out, not to the files as they sit here, so I cannot verify it locally.

## Deliberately not changed

- **`S1192` ×2** — a readable `VALUES` table of table-and-column pairs in PostgreSQL, and a dynamic SQL fragment in T-SQL. Extracting constants would make both harder to read.
- **`S107` ×2** — `HeartBeatWorker` and Redis `AReceiveMessageQueryHandler`, both eight DI collaborators. A parameter object would exist only to get under the threshold.

Both to be accepted with reasons once the rescan shows what is left.

## Verification

All eleven unit projects, zero failures: core 1226, RelationalDatabase 281, Dashboard.Api 222, Dashboard.Ui 273, LiteDb 182, Redis 212, SQLite 248, SqlServer 198, PostgreSQL 183, Memory 36, Dashboard.Client 92. Release build clean. The one converted assertion that lives in the shared integration project was exercised too.

No changelog entry — nothing here changes what a queue does.

## Summary by CodeRabbit

* **Chores**
  * Standardized SQL files to use LF line endings across platforms.
  * Reorganized internal member and method ordering without changing behavior.

* **Tests**
  * Updated collection, sequence, count, and containment assertions across unit and integration tests.
  * Preserved existing validation coverage and test behavior while aligning with current assertion patterns.

* **Documentation**
  * Adjusted documentation formatting without changing public APIs or functionality.


blehnen:
Good catch, and it was a real error rather than a style point. Pushed `330d89eb`.

I spotted that the analyzer's suggestion would lose order-independence, then picked the wrong replacement. `Assert.AreEquivalent` and `CollectionAssert.AreEquivalent` share a name and have **opposite** semantics — the old one compares contents in any order, the new one is a deep, order-sensitive structural comparison. The analyzer's own issue in `microsoft/testfx` names this as a migration trap and publishes the mapping:

```
CollectionAssert.AreEquivalent(a, b)  ->  Assert.AreSequenceEqual(a, b, SequenceOrder.InAnyOrder)
```

So the twelve sites in the previous commit were asserting an order none of them mean to assert — which tables a delete cleared, what a scheduler lists, what a dictionary holds.

**And every one of them passed.** Those collections do come back in a consistent order today, so nothing told me. That is precisely the failure mode I described in the PR body as the reason not to take the analyzer literally, and I introduced it anyway one step later.

All twelve now pass `SequenceOrder.InAnyOrder`. The arithmetic reconciles: 37 `AreSequenceEqual` calls total, 25 genuinely ordered because they came from `CollectionAssert.AreEqual`, 12 explicitly unordered. No `CollectionAssert` or `Assert.AreEquivalent` left in the tree.

Two corrections to what I wrote earlier in this PR:

- "`Assert.AreEquivalent` is the actual replacement and exists in 4.4" — wrong. It exists, but it is not the replacement.
- The twelve `(System.Collections.ICollection)` casts I cited as a bonus did come off, but via the wrong API. They stay off with the right one, and the two explicit type arguments went too: `AreSequenceEqual` takes `IEnumerable<T>` on both sides, so `IReadOnlyDictionary` and `IDictionary` infer a common `T` where `AreEquivalent` could not.

Re-verified: all eleven unit projects, 14 test runs, zero failures. Release build clean.

PR #364, opened 2026-09-17
TEMPORARY: show full Claude output to diagnose the review failure

blehnen:
**Do not merge.** This exists to read one error message, and will be closed once it has.

The Claude review job fails on its first model request:

```
model: claude-sonnet-5
num_turns: 1   duration_ms: 1828   modelUsage: {}   permission_denials_count: 0
is_error: true
```

No model usage recorded, so nothing reached a model. The secret is passed (`claude_code_oauth_token: ***`), and no tool permission was denied — which leaves a rejected token, a malformed secret value, or an entitlement problem, and the message is the only thing that separates them.

The action hides the SDK output by default, and debug logging does not lift it — the text stays in `claude-execution-output.json` on the runner, which is never uploaded. `show_full_output: true` is the only way to read it.

This PR exists so that output appears once, on a throwaway branch. Registered secrets remain masked by Actions and the code under review is public, but this is not a setting to leave on.

## Summary by Sourcery

CI:
- Temporarily enable full Claude action output in the review workflow to expose the hidden model-request error in build logs.

## Summary by CodeRabbit

* **Chores**
  * Temporarily expanded diagnostic output for automated code review workflows.
  * No changes to application functionality or the end-user experience.


blehnen:
Cannot work as designed. The action validates that the workflow file on the PR branch is identical to the version on the default branch and skips otherwise:

```
Skipping action due to workflow validation: Workflow validation failed. The workflow
file must exist and have identical content to the version on the default branch.
```

Any PR that edits `claude-review.yml` therefore guarantees a skip, so this branch can never produce the error it was opened to capture. It also explains why the earlier runs on #363 reported success without reviewing anything: they were workflow-editing PRs being skipped, not working reviews.

Worth recording: this is the upstream mitigation for the credential-theft concern raised on #363. A pull request author cannot modify the workflow to capture the token, because modifying it stops the action running at all.

To read the error, `show_full_output: true` has to be on the default branch while #361 re-runs.

PR #366, opened 2026-09-17
Temporarily show full Claude output to diagnose the review failure

blehnen:
Merge, read one error, revert. It does nothing else.

## Why this is needed

The Claude review fails identically every time it actually runs:

```
model: claude-sonnet-5
num_turns: 1   duration_ms: 1801   modelUsage: {}   permission_denials_count: 0
is_error: true
```

Nothing reaches a model. The secret is passed, no tool permission is denied, and the GitHub App token exchange over OIDC succeeds — that part of the log looks alarming but is working correctly.

It has now failed the same way with **two separately generated tokens**, so a mistyped or truncated secret is ruled out, as is a transient fault.

## Why it has to go to master

The action validates that the workflow file is identical to the version on the default branch and skips otherwise. That is a sound protection — it is what stops a pull request rewriting the workflow that holds the credential — but it also means this change cannot be tested from a branch. An earlier attempt (#364) was closed for exactly this reason.

## After merging

Re-run the Claude job on #365, read the message, then revert this. It prints every Claude message into a public build log; registered secrets stay masked and the code under review is public, but it should not stay on.

If you would rather not merge a diagnostic to master, the alternative is disabling the workflow until there is time to look properly — it currently reds every pull request, which is the kind of standing failure that trains you to ignore checks.

## Summary by Sourcery

Enhancements:
- Temporarily expose the Claude review action's full output to diagnose recurring model-request failures.


blehnen:
Superseded by #367, which disables the reviewer instead.

The diagnosis no longer needs a commit on master. `anthropics/claude-code-action#1614` documents this exact envelope, and the cause is already known from that thread: a 401 on the OAuth token for most reporters, a 429 for some. Merging a temporary `show_full_output` to the default branch would only confirm which of the two we have, and a direct token probe answers that without touching master.

Issue #362, opened 2026-09-17
Evaluate a replacement for qodo PR review (trial ended)


blehnen:
The qodo trial ended and this repository does not meet the 200-star bar for their OSS licence, so automated PR review is CodeRabbit only from now on. This issue is to decide whether that is enough and, if not, what to add.

## What we actually lost

Over #360 and #361, qodo found **one** thing of real value: a stale `connectionstring.txt` surviving in a persistent Jenkins workspace, which would have made the container work a silent no-op while every build stayed green. That was an excellent catch and neither CodeRabbit nor the author found it.

Against that, on #361 it raised three findings that were all false positives against deliberate repository conventions:

- integration test projects target `net10.0` only (CLAUDE.md documents this; Jenkins runs them on net10.0)
- `TreatWarningsAsErrors` is set on the 11 library projects and deliberately not on test projects
- XML documentation generation is likewise for shipped libraries, not test assemblies

All three fired on a `.csproj` whose entire diff was a single `PackageReference` line. So the replacement question is not "restore two bots" but "is there a second opinion that pays for the noise".

## Option: Claude Code GitHub Action

- https://github.com/anthropics/claude-code-action
- https://docs.claude.com/en/docs/claude-code/github-actions

Installs as a GitHub App and reviews PRs. Authenticates with either an Anthropic API key or a Claude subscription, so it needs the maintainer's account either way.

Usage is not a blocker here - a subscription covers it comfortably at this repository's rate, so the earlier concern about burning through credits does not apply.

The caveat that does stand is **correlated review**. Most changes here are authored with Claude. A Claude reviewer does not share the authoring session's context, but it does share the model's priors, so it is less likely to catch the class of thing a differently-built tool catches. CodeRabbit's value on #360 and #361 came precisely from reasoning differently.

That makes it worth adding as a **second pass** rather than as the primary reviewer: cheap to try, and a fresh-context read of a diff catches plenty even when it shares priors with the author.

## Other options to check

Free-for-open-source tiers, terms to be verified rather than assumed:

- **PR-Agent** (qodo's own open-source engine, self-hosted with your own LLM key) — https://github.com/qodo-ai/pr-agent
- **DeepSource** — https://deepsource.com
- **Sourcery** — https://sourcery.ai
- **Codacy** — https://www.codacy.com
- **Qodana** (JetBrains) — https://www.jetbrains.com/qodana/
- **Semgrep** — https://semgrep.dev
[...]

PR #367, opened 2026-09-17
Stop running the Claude review on pull requests (#362)

blehnen:
Turns off the Claude reviewer on pull requests. It stays in the repository and can still be started by hand, so a fix can be tested without exposing every pull request to it again.

## Why

Every run that reached a model died on its first request:

```
num_turns: 1   duration_ms: ~1800   modelUsage: {}   permission_denials_count: 0
is_error: true
```

The action surfaces that as a bare failure with no message.

This is **an open upstream problem, not our configuration**:

- [`anthropics/claude-code-action#1614`](https://github.com/anthropics/claude-code-action/issues/1614) — the same envelope exactly, with the real cause visible only under `show_full_output`: a **401** on the OAuth token for most reporters, a **429** for others. Reported on both Pro and Max, and reproduced on a pinned known-good action revision — so it is server side, not a release regression.
- [`anthropics/claude-code-action#1720`](https://github.com/anthropics/claude-code-action/issues/1720) — same envelope across many action versions, and with API-key auth as well as OAuth.

Re-minting the token does not help: reporters there went through four tokens against the same 401, and we saw the identical failure on two.

## Why not leave it red

A permanently failing check that no change of ours can fix is the fastest way to train everyone to stop reading checks — the same reason we removed a scanner false positive on #361 rather than living with it.

## Restoring it

Uncomment the `pull_request` trigger. The job condition keeps a `workflow_dispatch` clause because the fork check reads a pull-request payload a manual run does not have; a manual run requires write access regardless.

Worth a direct token probe before re-enabling, to tell a 401 from a 429 — the two causes behind an identical envelope want opposite responses.

## Summary by Sourcery

Enhancements:
- Disable automatic Claude reviews on pull requests while retaining manual workflow execution for testing and future recovery.

## Summary by CodeRabbit

* **Chores**
  * Automated pull request reviews are now triggered manually instead of running automatically.
  * Manual review runs remain available, while Dependabot-triggered reviews continue to be skipped.
  * Added documentation explaining the change in review workflow behavior.


blehnen:
Confirmed by direct probe: the token returns **401**, not 429, and a freshly generated one returns 401 as well.

That matches [`claude-code-action#1614`](https://github.com/anthropics/claude-code-action/issues/1614) exactly, including the detail that re-minting does not help - reporters there went through four tokens for the same result. So this is the OAuth rejection rather than rate limiting, it is server side, and nothing in this repository can fix it.

Disabling is the right call until that issue closes. Worth watching it for a fix, and re-probing the token before re-enabling.

